PR #171, opened 2015-01-01
Add error and status parsing from ES exception

ebuildy:
Since ElasticSearch throws JSON encoded exceptions, I think it's a good idea to expose the real error and status instead of the raw message. We could also consider to have a ElasticsearchException class that all exceptions could override.


polyfractal:
I like it.  Could you:

1. Sign the [Elasticsearch CLA](http://www.elasticsearch.org/contributor-agreement/) so that it can be merged
2. Add some tests, particularly with input that isn't JSON encoded?  ES quite often spews non-JSON exceptions (particularly stack traces, but sometimes for 400 exceptions too).  Your code looks like it will handle it, but a test would be good to verify.

polyfractal:
Closing, since ES now throws real structured JSON exceptions instead of quasi-structured stuff from before 2.0.

PR #176, opened 2015-01-12
Rework JSON Serializer to Throw Useful Exceptions

### LINKED ISSUE #175: Throw useful exception when JSON serialization or deserialization fails
I have an application right now where I'm seeing a lot of errors like this:

```
Elasticsearch\Common\Exceptions\BadRequest400Exception -- {"error":"RemoteTransportException[[ip-10-0-3-88][inet[/10.0.3.88:9300]][indices:data/write/index]]; nested: MapperParsingException[failed to parse, document is empty]; ","status":400}
```

It's happening because serialization of the request body is failing (I know this because I tried to log a that body with `json_encode` and showing nothing).

I think this library should check [`json_last_error`](http://php.net/manual/en/function.json-last-error.php) in its [serializers](https://github.com/elasticsearch/elasticsearch-php/tree/master/src/Elasticsearch/Serializers) and throw a meaningful exception about the serialization failure.

I'd be happy to submit a pull request for this if you'd like.

chrisguitarguy:
Closes #175

Previously `SmartSerializer` was the only class that sometimes threw
exceptions when something went wrong during JSON deserialization. This
changes that so all serializers throw.

- `JsonSerializationError` is thrown when something goes wrong during
  `json_encode`
- `JsonDeserializationError` is thrown when something goes wrong during
  `json_decode`

Both are subclasses of `JsonErrorException`, so anyone already catching
that should be able to continue doing so.

There's a also a bit of restructuring here and some additional tests to
cover some more paths in the serializers.

### DIFF
--- a/src/Elasticsearch/Common/Exceptions/Serializer/JsonDeserializationError.php
+++ b/src/Elasticsearch/Common/Exceptions/Serializer/JsonDeserializationError.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Elasticsearch\Common\Exceptions\Serializer;
+
+/**
+ * Thrown when there's an error unserializing a JSON string to a PHP array.
+ *
+ * @author   Christopher Davis <cdavis9999@gmail.com>
+ * @license  http://www.apache.org/licenses/LICENSE-2.0 Apache2
+ * @link     http://elasticsearch.org
+ */
+final class JsonDeserializationError extends JsonErrorException
+{
+    // noop
+}
--- a/src/Elasticsearch/Common/Exceptions/Serializer/JsonSerializationError.php
+++ b/src/Elasticsearch/Common/Exceptions/Serializer/JsonSerializationError.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Elasticsearch\Common\Exceptions\Serializer;
+
+/**
+ * Thrown when there's an error serialization a PHP object or array to a JSON
+ * string.
+ *
+ * @author   Christopher Davis <cdavis9999@gmail.com>
+ * @license  http://www.apache.org/licenses/LICENSE-2.0 Apache2
+ * @link     http://elasticsearch.org
+ */
+final class JsonSerializationError extends JsonErrorException
+{
+    // noop
+}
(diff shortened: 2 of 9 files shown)


chrisguitarguy:
I already signed the [CLA](http://www.elasticsearch.org/contributor-agreement/). Should be attached to my github username?

fprochazka:
As I've already mentioned somwhere in this repo... such solution is insufficient 

http://phpfashion.com/how-to-encode-and-decode-json-in-php

chrisguitarguy:
I would argue that your solution, while great for a framework, is a bit overkill for a library. The goal is to provide some sort of useful exception feedback when something goes wrong. If your specific use case requires covering all of those edge causes the correct solution is to roll your own serializer. Someone using [PECL jsonc](http://pecl.php.net/package/jsonc) should definitely roll their own serializer to deal with its specifics.

Library's shouldn't try to solve all possible problems, only supply reasonable defaults and make things extensible.

chrisguitarguy:
It's also worth looking at Symfony's [serializers](https://github.com/symfony/symfony/tree/master/src/Symfony/Component/Serializer/Encoder) which are *very* thin wrappers around `json_{encode,decode}`.  Zends are [as well](https://github.com/zendframework/zf2/blob/master/library/Zend/Serializer/Adapter/Json.php) when their [custom encoders/decoders](https://github.com/zendframework/zf2/tree/master/library/Zend/Json) aren't in use.

polyfractal:
FYI, traveling...will get to this as soon as I get back home.  I'm wary to merge it on my travel laptop, want to run it through the test suite first :)

Thanks!

Re: Nette...I definitely want to pull some of their json logic in for 2.0, but we don't need the whole set of conditions.  IIRC they support down to some really low version of PHP (even less than 5.3.9 that we currently support), and with 2.0 bumping the min version to 5.4, we can drop even more edge-cases.

So yeah, on the to-do list...I just don't want to tackle it yet :)

polyfractal:
Merged!  Really sorry for the delay on this, I got caught up with ES-PHP 2.0 and Elasticon stuff.  Going to let this bake on CI for a bit before tagging.  

Thanks again!

PR #184, opened 2015-01-21
This parameter doesn't exist anymore

mdzor:


Bargok:
When I look into master source `Elasticsearch/Client.php` (https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Client.php#L1570) I still see:

```php
$handler = new StreamHandler(
    $this->params['logPath'],
    $this->params['logLevel'],
    $this->params['logBubble'],
    $this->params['logPermission']
);
```

PR #185, opened 2015-01-21
LogLevel is using Monolog standard

mdzor:
Psr\Log\LogLevel string levels are not supported anymore

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -129,14 +129,15 @@ $client = new Elasticsearch\Client($params);
 ----
 {zwsp} +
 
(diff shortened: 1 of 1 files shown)


polyfractal:
Hmm?  I believe PSR Log is still supported...it was added in 1.11 because of an issue we found via ES-PHP (https://github.com/elasticsearch/elasticsearch-php/issues/125, which resulted in: https://github.com/Seldaek/monolog/commit/feb19af2d42ab598e3e5e77e8f4bd51b82538acb)

The monolog readme says:

> As of 1.11.0 Monolog public APIs will also accept PSR-3 log levels.

PR #190, opened 2015-02-12
allow pimple/pimple:~2.0

mlebkowski:
Changed `pimple/pimple` dependency to allow `~2.0` also.

There are no code changes [between `pimple/pimple:~2.0` and `pimple/pimple:~3.0`](https://github.com/silexphp/Pimple/compare/v2.1.1...v3.0.0) so `~2.0` would be good enough for any project and it would remove a conflict with `silex/silex:~1.0`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
   "require": {
     "php": ">=5.3.9",
     "monolog/monolog": "~1.11",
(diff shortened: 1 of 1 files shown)


mlebkowski:
nevermind :o

PR #194, opened 2015-03-03
Fix typo in 'metric' argument of Stats function in IndicesNamespace.php

iJackUA:
Argument was not correctly extracted, so it was impossible to compose requests like `/_stats/indices` by setting `metric` param
Also related to 2.0 and master

### DIFF
--- a/src/Elasticsearch/Namespaces/IndicesNamespace.php
+++ b/src/Elasticsearch/Namespaces/IndicesNamespace.php
@@ -232,7 +232,7 @@ class IndicesNamespace extends AbstractNamespace
      */
     public function stats($params = array())
     {
-        $metric = $this->extractArgument($params, '$metric');
+        $metric = $this->extractArgument($params, 'metric');
 
         $index = $this->extractArgument($params, 'index');
 


polyfractal:
Arg, typing is hard apparently.  Thanks for the fix!

Could you sign the [Elasticsearch CLA](http://www.elasticsearch.org/contributor-agreement/) so I can merge this in?

iJackUA:
Sure, done!
https://elasticsearch.echosign.com/public/viewAgreement?tsid=2AAABLblqZhBUO-j1fenvS9eF4eMVrj6BdZP44VfKd-2VHbZHGDqB6jpdtZ8HIFIvE3dlW8lVbQw*&

polyfractal:
Great, thanks!  Merged!

polyfractal:
Heh, as an aside, this also exposed a bug that was being silently ignored by the tests before (since the metric wasn't extracted, all the metrics were being returned and it passed the tests).  Thanks!

Fixed in https://github.com/elasticsearch/elasticsearch-php/commit/756161b2256aa62ceabac27eb9fa2ce539098dc8

PR #198, opened 2015-03-09
Composer: PSR-4 + autoload-dev

marc-mabe:


polyfractal:
Sorry for the delay on this...slipped through my inbox without noticing.  Are there any gotchas to PSR-4 that I should be aware of?  I haven't really followed it, and don't know much about it.  It looks nice from a superficial skimming...  From what I can gather, it's mostly dropping backwards compatibility features that were inherited from PEAR?

Also, would you mind signing the [Elasticsearch CLA](https://www.elastic.co/contributor-agreement/) so I can merge this?

marc-mabe:
PSR-4 is compatible with PSR-0 if you don't use underscores as namespace separator (this was the standard was before PHP has namespaces). You also could remove the directories that only exist for pointing into the first namespace but this would require changing directory layout and would be a BC break. Without changing the directory structure or adding non namespaces classes including an underscore there is no BC break.
-> Btw. PSR-0 is deprecated - see http://www.php-fig.org/psr/psr-0/ ;)

The `autoload-dev` section is important for optimizing composer autoloader for production as this autoload configuration is only included by composer on development environment and dependencies are ignored. -> This drastically reduces the generated classmap file.

This PR doesn't contain license specific logic. It's only a composer configuration. I'm ok with it if you or someone else creates a new PR/commit without referencing me.
-> No CLA for  this one ;)

polyfractal:
Awesome, thanks for the explanation.  And good to know about the deprecation, I didn't realize that had happened.  Will definitely update to PSR-4!  (especially since it doesn't seem like it will require any real changes to the project)

We're required to get CLA signatures, even for small things like typos in documentation or configuration changes.  Since you're ok with it, I'll just apply the change myself (but I'll leave a note of attribution in the commit message).

Thanks again!

polyfractal:
Merged in the change myself, thanks again!

PR #206, opened 2015-03-30
Fiz Guzzle usage

davividal:
Some methods require 2, 3 or 4 parameters. Always passing 4 will break
things.

### DIFF
--- a/src/Elasticsearch/Connections/GuzzleConnection.php
+++ b/src/Elasticsearch/Connections/GuzzleConnection.php
@@ -153,31 +153,49 @@ class GuzzleConnection extends AbstractConnection implements ConnectionInterface
             $method = 'POST';
         }
 
+        $headers = array();
         if (isset($body) === true) {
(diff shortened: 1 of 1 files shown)


polyfractal:
Hi there, thanks for the PR!  Could you perhaps show an example where it breaks?  Elasticsearch doesn't use PATCH, OPTIONS, etc, so I'm a bit confused what's breaking.  The integration test suite runs through most of the ES APIs and it's never broken because of Guzzle parameters either.

Also, it looks like some of the HTTP basic auth code was dropped.  That will need to stay so as to support anyone using an authorization proxy, or Shield.

davividal:
When you try to use a HEAD request, it will break.

I was using it through Client::indices()->create(). It was breaking beacause of auth. Looking closely, I was using ~1.0.

I'll compare it against v1.3.3.

polyfractal:
closed in favor of #207

PR #207, opened 2015-03-30
Fix guzzle client

davividal:
When trying to use a HEAD request, it will break Basic auth.

I think we should close the other merge request. I opened it against the master branch, which was a mistake, sorry.

This is the correct branch.

### DIFF
--- a/src/Elasticsearch/Connections/GuzzleConnection.php
+++ b/src/Elasticsearch/Connections/GuzzleConnection.php
@@ -153,25 +153,49 @@ class GuzzleConnection extends AbstractConnection implements ConnectionInterface
             $method = 'POST';
         }
 
+        $headers = array();
         if (isset($body) === true) {
-            $this->lastRequest = array( 'request' => array(
+            $headers['content-type'] = 'application/json';
+        }
+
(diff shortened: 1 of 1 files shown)


polyfractal:
So I was confused why the auth parameters were showing in the other PR's diff, but not this one.  As it turns out commit f3bdf05103ac037a37b6cd2160abb97952bf8f2e was never cherry-picked into 1.0 ... it only existed in master.

I just cp'd it into 1.0 and pushed.  Could you try a test with either dev-master or 1.0.x-dev to see if it fixes your issues?  I think this is an inline auth bug rather than a problem with Guzzle.

Also, could you add the error message that you are receiving, and the commands you are running so I can test locally?

polyfractal:
I tagged `v1.3.4` last night, which includes this fix.  Please upgrade and see if that fixes your issue?

davividal:
@polyfractal it worked! Thank you very much.

polyfractal:
Great!  Happy to hear, and sorry again for missing that commit.  It was sitting on master for ages without being merged into 1.0...argh :)

PR #209, opened 2015-04-02
Add JSON_FORCE_OBJECT

ebuildy:
Sometimes ES API doesnt accept [] but only {}. For instance, in aggregation to use global scope.

JSON_FORCE_OBJECT is available from Php 5.3.0

Maybe this should be a parameter?

### DIFF
--- a/src/Elasticsearch/Serializers/AbstractJsonSerializer.php
+++ b/src/Elasticsearch/Serializers/AbstractJsonSerializer.php
@@ -23,7 +23,7 @@ abstract class AbstractJsonSerializer implements SerializerInterface
      */
     protected static function jsonEncode($value)
     {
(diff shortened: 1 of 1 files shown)


ebuildy:
Oups, there are some side effects :/

So we need to replace void [] by {} after the JSON serialization :/

polyfractal:
Yeah, there's no elegant way to autodetect this sort of thing, unfortunately.  If you look at the JSON serialization code, there are a few hacks to help...but nothing that works across all the APIs.

The robust solution is to use empty PHP objects (`new \StdClass()`).  See: http://www.elastic.co/guide/en/elasticsearch/client/php-api/current/_dealing_with_json_arrays_and_objects_in_php.html

PR #213, opened 2015-04-05
Exists Endpoint Tweak

mhwest13:
This is to allow exists calls to be made on indexes, indexes and types, as well as documents like before.

### DIFF
--- a/src/Elasticsearch/Endpoints/Exists.php
+++ b/src/Elasticsearch/Endpoints/Exists.php
@@ -28,28 +28,20 @@ class Exists extends AbstractEndpoint
      */
     protected function getURI()
     {
-        if (isset($this->id) !== true) {
-            throw new Exceptions\RuntimeException(
-                'id is required for Exists'
-            );
-        }
         if (isset($this->index) !== true) {
             throw new Exceptions\RuntimeException(
                 'index is required for Exists'
             );
         }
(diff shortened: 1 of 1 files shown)


polyfractal:
Hi, thanks for the PR!  However, this functionality already exists under the `indices` namespace.  E.g.:

```php
$indexExists = $client->indices()->exists(['index' => 'my_index']);
$typeExists = $client->indices()->existsType(['index' => 'my_index', 'type' => 'my_type']);

// Indices namespace also contains template and alias exists:
$templateExists = $client->indices()->existsTemplate(['name' => 'my_template']);
$aliasExists = $client->indices()->exists(['index' => 'my_index', 'name' => 'my_alias']);
```

The way the client is split is that the "root" namespace is reserved for common document-oriented operations, while the rest of the operations (indices, cluster, snapshot/restore, etc) are segmented into respective modules.  This keeps the root namespace from getting too cluttered, and prevents APIs from taking on too much responsibility.

It also helps prevent silly accidents, like a `delete` removing an entire index, when you meant to delete a single document but forgot to include an index name.  By segmenting functionality, you have to "opt in" to non-document operations.

mhwest13:
Thanks for the pointer info, I'll definitely switch over to using that in my code. I had missed that earlier while pouring through all of the API / Endpoints / possibilities in the docs and code.

Cheers

PR #216, opened 2015-04-13
Fix Fatal Errors when using ClientBuilder::defaultLogger()

simplechris:
When enabling the logger as described [in the docs](http://www.elastic.co/guide/en/elasticsearch/client/php-api/2.0/_configuration.html#enabling_logger), Fatal Errors are thrown.

This seems to be due to missing `use` statements for the following classes used within the `ClientBuilder::defaultLogger()` method

-  `Monolog\Logger`
-  `Monolog\Handler\StreamHandler`
-  `Monolog\Processor\IntrospectionProcessor`

This patch adds the missing `use` statements for these classes.


polyfractal:
Doh, thanks for catching this.  What a silly mistake :P

Issue #201, opened 2015-03-19
CurlMultiHandler with PHP 5.4 [v2.0.0-beta1]


simplechris:
I'm trying out the new `v2.0.0-beta1`, but I'm having some trouble with the [async/future mode](http://www.elastic.co/guide/en/elasticsearch/client/php-api/2.0/_future_mode.html), using with PHP 5.4. The problem seems to be with `Elasticsearch\ClientBuilder::multiHandler()`

```php
<?php

require 'vendor/autoload.php';

// Build Client instance
$client = Elasticsearch\ClientBuilder::create()
    ->setHosts(['localhost:9200'])
    ->setHandler(Elasticsearch\ClientBuilder::multiHandler())
    ->build();

```

**Fatal error: Uncaught exception 'RuntimeException' with message 'CurlMulti handler requires cURL.' in /chris/code/playground/elasticsearch-php-2-0-0-beta1-testing/vendor/elasticsearch/elasticsearch/src/Elasticsearch/ClientBuilder.php on line 117**

Here's my setup: 

```
$ php -i

PHP Version => 5.4.37
...
curl

cURL support => enabled
cURL Information => 7.37.1
Age => 3
Features
AsynchDNS => Yes
Debug => No
GSS-Negotiate => Yes
IDN => No
IPv6 => Yes
Largefile => Yes
NTLM => Yes
SPNEGO => No
SSL => Yes
SSPI => No
krb4 => No
libz => Yes
CharConv => No
Protocols => dict, file, ftp, ftps, gopher, http, https, imap, imaps, ldap, ldaps, pop3, pop3s, rtsp, smtp, smtps, telnet, tftp
Host => x86_64-apple-darwin14.0
SSL Version => SecureTransport
ZLib Version => 1.2.5

```

so, when running PHP 5.4 with cURL, the above exception gets thrown I think due to [this check for `curl_reset`](https://github.com/elastic/elasticsearch-php/blob/0e57e514c63a29a836607f30d401b93180da3ba9/src/Elasticsearch/ClientBuilder.php#L114) which the [documentation](http://php.net/manual/en/function.curl-reset.php) states to be a >= PHP 5.5 feature. Before I provide a patch, I wanted to check if there was a particular reason to check for `curl_reset` and not `curl_multi_init`?

polyfractal:
[...]

PR #220, opened 2015-04-16
ClientBuilder fixes for PHP 5.4

simplechris:
Leading on from https://github.com/elastic/elasticsearch-php/issues/201#issuecomment-93821595

- Upstream `GuzzleHttp\Ring\Client\CurlHandler` requires `curl_reset` (available from PHP 5.5) so when using `ClientBuilder::singleHandler()` directly, the check for `curl_reset` should take place or throw the Exception. 
- `ClientBuilder::defaultHandler()` should also perform this check, falling back to `GuzzleHttp\Ring\Client\CurlMultiHandler` in environments where `curl_reset` is not available (the multi handler has no such dependency on `curl_reset`)
- The `ClientBuilder::multiHandler(`) method doesn't require `curl_reset`, but does require `curl_multi_init`, which should be available to those with the curl extension enabled.

Ref:  https://github.com/elastic/elasticsearch-php/issues/201

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -95,7 +95,7 @@ class ClientBuilder
         $future = null;
         if (extension_loaded('curl')) {
             $config = array_merge([ 'mh' => curl_multi_init() ], $multiParams);
-            if (function_exists('curl_multi_init')) {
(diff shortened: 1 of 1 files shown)


polyfractal:
Thanks for doing the legwork on this, really appreciate it!

PR #222, opened 2015-04-20
travis: recent PHP added

TomasVotruba:
- `--dev` is by default for almost a year

Further, I have some suggestions to improve build process.
No need to test matrix every PHP version x every ES_VERSION. Same for coverage.


TomasVotruba:
Closing in favor of #224

Issue #103, opened 2014-07-02
The travis setup looks broken


stof:
The repo contains a config file for Travis, but the last build on Travis was 5 months ago: https://travis-ci.org/elasticsearch/elasticsearch-php/builds
It looks like the hook triggering builds is missing

TomasVotruba:
@polyfractal Looks like almost year passed and CI is here.

I can take care of Travis (got many issues at from the start, so I understand your attitude).
Since there are unit tests and that's all I need to make it run.

polyfractal:
Thanks for the poke.  We opened Jenkins a while ago but I completely forgot to make note of it here.

Our CI summary page:  http://build.elastic.co/
ES-PHP 1.0: http://build-eu-00.elastic.co/job/es-php_core_10/
ES-PHP Master: http://build-eu-00.elasticsearch.org/view/Clients/job/es-php_core/
ES-PHP Integration (currently unstable due to a jenkins issue): http://build-eu-00.elasticsearch.org/view/Clients/job/es-php_integration/

I'll add build badges to the README soon.

I stand by my statement from before:  I'm disinclined to invest energy in Travis since we already have a (relatively) robust CI setup.  It certainly isn't without it's faults, but considering the complexity testing all the various versions, Travis is still going to be a pain.

stof:
@polyfractal maybe Travis could be kept with testing against a single ES version (the one preinstalled on Travis, to simplify the setup), which would allow to have tests running on PRs for a faster feedback loop rather than having tests running only after a merge. Having tests against a single version of ES for pending PRs is still better than no tests running at all

polyfractal:
@stof Hmm, yeah, that seems fair.  It would provide a simple litmus test if the PR is completely broken or not.  And it would be less work for Travis too, I would feel a little bad about running the full matrix on their infrastructure.

Might as well give it a shot...worst case we can just disable Travis again if it proves too much hassle, or not helpful enough.  I'll lodge another request with our infra team to investigate Jenkins-PR plugins...I think that request was lost in an internal overhaul of the jenkins system.

:+1: from me for a smaller version of Travis running :)

TomasVotruba:
@polyfractal  Great job! :+1: 

Let me handle any hassles you came across.

polyfractal:
Closing, travis support was added a few weeks ago.

(currently broken, but a totally different problem.  Will fix that soon!)

PR #224, opened 2015-04-20
[WIP] Tests ci

TomasVotruba:
Todo:

- [x] wait till Travis is up again
- [ ] fix issues that will show up


fprochazka:
:+1:

polyfractal:
Following up on #103 discussion:

- Thoughts about testing against nightly snapshots?  I don't (personally) want to update the travis.yml file every time a new version of ES is released.  We could grab nightly snapshot builds that are used in Jenkins.  (e.g. http://s3-eu-west-1.amazonaws.com/build-eu.elasticsearch.org/origin/1.4/nightly/JDK7/elasticsearch-latest-SNAPSHOT.zip for `1.4.x`)

- It looks like this will run the `YamlRunnerTest` as well as all the regular unit tests.  We need to either exclude this test or set `TEST_BUILD_REF` env variable and invoke `RestSpecRunner` for each version of Elasticsearch.  This script will checkout the relevant YAML integration tests (which change from version to version).  Otherwise the integration tests will break if the wrong yaml suite is used

.travis.yml
```diff
@@ -1,21 +1,47 @@
...
-  - ES_VERSION=0.90.1 ES_TEST_HOST=http://localhost:9200
-  - ES_VERSION=0.90.2 ES_TEST_HOST=http://localhost:9200
-  - ES_VERSION=0.90.3 ES_TEST_HOST=http://localhost:9200
-  - ES_VERSION=0.90.4 ES_TEST_HOST=http://localhost:9200
-  - ES_VERSION=0.90.5 ES_TEST_HOST=http://localhost:9200
+  - ES_VERSION="1.5.1"
+  - ES_TEST_HOST=http://localhost:9200
+
+before_install:
+  - curl -O https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-$ES_VERSION.deb && sudo dpkg -i --force-confnew elasticsearch-$ES_VERSION.deb
```
stof: why not just using the ES service available on Travis ?

.travis.yml
```diff
@@ -1,21 +1,47 @@
...
+
 env:
-  - ES_VERSION=0.90.0 ES_TEST_HOST=http://localhost:9200
-  - ES_VERSION=0.90.1 ES_TEST_HOST=http://localhost:9200
-  - ES_VERSION=0.90.2 ES_TEST_HOST=http://localhost:9200
-  - ES_VERSION=0.90.3 ES_TEST_HOST=http://localhost:9200
-  - ES_VERSION=0.90.4 ES_TEST_HOST=http://localhost:9200
-  - ES_VERSION=0.90.5 ES_TEST_HOST=http://localhost:9200
+  - ES_VERSION="1.5.1"
+  - ES_TEST_HOST=http://localhost:9200
```
stof: this should use the ``env: global:`` config to avoid creating multiple jobs (the short syntax is for env-based matrix)

.travis.yml
polyfractal: Does Travis state what version is used by their service?  And/or how often it get's update?  Not really a problem, we just need to know what version so we can grab the relevant YAML tests.

I don't see anything in the docs (http://docs.travis-ci.com/user/database-setup/#ElasticSearch)

stof: They currently use the official Debian package for 1.4.0 (the version being used can be found in https://github.com/travis-ci/travis-cookbooks/blob/master/ci_environment/elasticsearch/attributes/default.rb)

Any update to newer versions would then be announced on http://docs.travis-ci.com/user/build-environment-updates/
The latest ES update was 5 months ago: http://docs.travis-ci.com/user/build-environment-updates/2014-12-09/ They haven't updated it to newer versions since then in following updates (it could be requested though). Env updates will only happen every 2 months anyway

Regarding getting the version, isn't there a way to get the version of ES from the installed ES itself (something like ``elasticsearch --version`` ?) If no, you would have to update the version manually after Travis updates impacting elasticsearch

polyfractal: Gotcha, thanks for the info.

We can get the version from ES...it just complicates things a bit.  There's no way to get it from the executable itself (unfortunately).  We'd have to spin up the server first, then do a `curl -XGET localhost:9200/` to get the version number. Then checkout the appropriate yaml test branch.

TomasVotruba:
@polyfractal Do you plan to active Travis builds, so I can easily debug all issues?

polyfractal:
Yes, sorry...been super busy with a dozen different projects.  I'll make this a priority to get configured tomorrow.  Thanks for your patience!

polyfractal:
@TomasVotruba Thanks for your patience, I've been in/out of town for work.

I pushed a new [`travis` branch](https://github.com/elastic/elasticsearch-php/tree/travis) which is a mirror of `master` with your PR applied.  I want to keep the travis work separate from the normal branches, just in case transitory changes break the normal builds.  I modified the `.travis.yml` to only build the `travis` branch, and have enabled the webhook.

Let me know what else you need moving forward :)

TomasVotruba:
@polyfractal Work is important. I'm glad you found some time.

I need you to [enable the repository](https://travis-ci.org/elasticsearch/elasticsearch-php) on Travis.
Just login and on the left is "My repositories". There should be "+" for this.

Let me know if you come across any troubles.

polyfractal:
Hehe, well, maintaining this repo is part of my work too :)  I feel like I'm juggling geese some days.

I believe the repo is already enabled at Travis:
screen shot 2015-05-22 at 10 59 04 am

Although it looks like I didn't configure the `branch` directive correctly, since it appears to be running on master and not just travis:  https://travis-ci.org/elastic/elasticsearch-php/builds/63662779

stof:
@polyfractal Travis reads the config file in the commit being tested. The travis branch does not contain the ``branch`` directive

polyfractal:
@stof Ahh, so I need to [add this](https://github.com/elastic/elasticsearch-php/blob/travis/.travis.yml#L14-L16) to all the branches, not just the travis branch?

That makes sense, Travis would only have the .travis.yml for the committed branch...it isn't going to pull a separate branch to check.

TomasVotruba:
Just a refresh, what needs to be done here to merge?

polyfractal:
Thanks for the poke.  I need to do some git shenanigans related to the 2.0 release first*, then I'll merge this into master and we can iterate from there.  Plan is to get this stuff squared away by the end of the week at the latest, preferably today/tomorrow.

* *2.0 originally started as an experimental branch, then became a real branch...but master is still on the 1.0 codebase. Need to fix that in the least irritating way first*

TomasVotruba:
Rather merge #286, this is very outdated and shall be closed.

PR #229, opened 2015-04-29
Readme: simplify install

TomasVotruba:
Composer will select last stable version by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,29 +67,19 @@ Installation via Composer
 -------------------------
 The recommended method to install _Elasticsearch-PHP_ is through [Composer](http://getcomposer.org).
 
-1. Add ``elasticsearch/elasticsearch`` as a dependency in your project's ``composer.json`` file (change version to suit your version of Elasticsearch):
-
-    ```json
-        {
-            "require": {
-                "elasticsearch/elasticsearch": "~1.0"
-            }
-        }
-    ```
-
-2. Download and install Composer:
+1. Download and install Composer:
 
     ```bash
         curl -s http://getcomposer.org/installer | php
     ```
 
-3. Install your dependencies:
+2. Install last stable version:
 
     ```bash
-        php composer.phar install --no-dev
+        php composer.phar require elasticsearch/elasticsearch
     ```
 
-4. Require Composer's autoloader
+3. Require Composer's autoloader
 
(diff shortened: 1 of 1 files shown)


polyfractal:
I like this, especially since I want to start pushing people towards `v2.0` once it is out of beta.

Would you mind updating the docs to reflect this simplification too?

Also, would you mind signing the [Elasticsearch CLA](http://www.elasticsearch.org/contributor-agreement/) so I can merge this?

TomasVotruba:
@polyfractal CLA signed.

Ad doc, seems like there's lot of [duplicated info](https://github.com/elastic/elasticsearch-php/blob/master/docs/installation.asciidoc#version-matrix) that can be found in [Readme](https://github.com/elastic/elasticsearch-php#version-matrix), which is outdated for this reason.

I'd recommend not to duplicate any content, and keep installation in Readme, since it's very easy step.

polyfractal:
Yeah, I agree, it would be nice to only have one set of instructions instead of the same content duplicated.  Unfortunately, a fair number of people land on the documentation from search engines, and see the [site documentation](http://www.elastic.co/guide/en/elasticsearch/client/php-api/current/_installation_2.html) before the project Readme.  It would be odd if they couldn't find the installation in the regular docs.

TomasVotruba:
I agree. It's common practice to have link in Readme to full docs, e.g. [DoctrineBundle](https://github.com/doctrine/DoctrineBundle#documentation).

No duplication and everything could be found.

TomasVotruba:
I'd suggest link Readme to [docs install](https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/_installation_2.html)

polyfractal:
Circling back on this.  As much as I like DRY, I also like having installation + small code examples in the repo just so people can see a flavor.  It's only minor duplication with the docs.  And all the other ES repos do similar, so it is best if we stay semi-consistent with the other clients:

https://github.com/elastic/elasticsearch-py
https://github.com/elastic/elasticsearch-ruby
https://github.com/elastic/elasticsearch-js

If you don't want to update the docs, I'll merge this and do it myself.  Thanks! :)

TomasVotruba:
> If you don't want to update the docs, I'll merge this and do it myself. Thanks! :)

That would be great, I'm out off time now. Thank you

polyfractal:
Merged in https://github.com/elastic/elasticsearch-php/commit/dc8e072909af8ea34dfcb0bf01ee7d5d87469dea

PR #254, opened 2015-06-21
Update README to link to Composer site for how to install Composer.

### LINKED ISSUE #251: Better/more secure way to install with composer?
Currently in our documentation, we recommend people [install the PHP client with composer](https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/_quickstart.html) by piping a downloaded script into PHP.  That seems bad.  The download is not secured by HTTPS and there is no verification of the downloaded script.  Even if not done as an admin account, this seems like a bad idea. Is there no better way to recommend installing the client?

joshuar:
Super simple documentation patch to refer users to the Composer website for how to install Composer.  This means that if Composer changes their installation method in the future, our documentation does not need to be changed (assuming the URLs stay the same). This also fixes #251 in a way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,11 +77,7 @@ The recommended method to install _Elasticsearch-PHP_ is through [Composer](http
         }
     ```
 
-2. Download and install Composer:
-
-    ```bash
-        curl -s http://getcomposer.org/installer | php
-    ```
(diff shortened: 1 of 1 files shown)


polyfractal:
++  I'll merge this later today or first thing tomorrow.  Need to prep for a webinar, then the rest of the week is PHP fixes :)

Thanks!

polyfractal:
Merged in https://github.com/elastic/elasticsearch-php/commit/7cb331f01772ee0d090b2dd5536dffbd42cc954a

PR #267, opened 2015-07-15
remove unused code

Dyachenko:
Remove unused code from Endpoints namespace

### DIFF
--- a/src/Elasticsearch/Endpoints/Bulk.php
+++ b/src/Elasticsearch/Endpoints/Bulk.php
@@ -37,9 +37,6 @@ class Bulk extends AbstractEndpoint implements BulkEndpointInterface
      */
     public function setBody($body)
     {
-        if (isset($body) !== true) {
-            return $this;
-        }
 
         if (is_array($body) === true) {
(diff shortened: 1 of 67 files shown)


polyfractal:
Thanks for the PR, deleting code is always a good thing!

Some of these are definitely good, there is a fair amount of leftover cruft from the original autogeneration.  Mostly the stuff involving setting the `$uri` unnecessarily a second time.

The various checks for `isset($var) !== true` I'm less sure about.  It's still early in the morning and I haven't had enough :coffee: yet, but I feel like there are edge cases that these are catching.  But I agree that a lot of them *seem* useless, since if a user set NULL the variable will remain NULL regardless of the `isset`

I'll work up some edge case tests to convince myself these are fine.  It's probably all good and I'm just worrying for no reason :)

polyfractal:
Just a note that I haven't forgotten about this, and am going to try and clear some time to sit down and merge this.  Thanks for being patient :)

polyfractal:
Closing, this is just too big to review and too much code has changed in the mean time.  Sorry :(

PR #270, opened 2015-07-28
comoposer command error

jianfengye:
comoposer command error


polyfractal:
Indeed you are right.  I think I wrote those docs optimistically, thinking the beta would breeze by.  Whoops. :)

Would you mind signing the [Elasticsearch CLA](http://www.elasticsearch.org/contributor-agreement/) so that I can get this merged?

Issue #258, opened 2015-06-29
Abstracting scrolled responses with PHP Iterators


arturom:
Currently, it's possible to [scroll through search results](https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/_search_operations.html#_scan_scroll) using `$client->search()` and then `$client->scroll()` 

This pattern has come up repeatedly for me to export data in a background process. It would be nice to have a way to abstract it. A  solution would be to use PHP's [Iterator interface](http://www.php.net/manual/en/class.iterator.php). I've come up with two separate iterator classes: One iterator would be used iterate over every "page" of results and one to iterate over every individual hit.

```php
$client = new Elasticsearch\Client($config);

$search_params = array(
    'search_type' => 'scan',
    'scroll'      => '5m',
    'index'       => 'twitter',
    'size'        => 1000,
    'body'        => array(
        'query' => array(
            'match_all' => new StdClass
        )
    )
);

$pages = new Elasticsearch\Iterators\PageIterator($client, $search_params);
$hits = new Elasticsearch\Iterators\HitIterator($pages);

// Sample usage of iterating over hits
foreach($hits as $hit) {
    // do something with this hit. For example: write to a CSV, update a database, etc..
    echo $hit['_id'], PHP_EOL;
}
// Sample usage of iterating over scrolled responses
foreach($pages as $page) {
    // do something with this hit. For example: copy its data to another index, etc...
    echo count($page['hits']['hits']), PHP_EOL;
}

```
I have an implementation of these classes in my forked repo: https://github.com/arturom/elasticsearch-php/commits/iterators

Do you think these classes have a place in this library?

polyfractal:
Wow, this totally slipped through my inbox.  Sorry for the radio silence!

Firstly, I really do like this approach; it fits perfectly with an iterator.  Unfortunately, it cannot be included *directly* in Client for two reasons:  1.0 supports down to PHP 5.3.9, and the client is supposed to be an "un-opinionated" client.  Something like an iterator starts to make opinions about how someone should consume the ES API, which is against the design goals.

BUT, for a long time I've wanted to add a "helper" class that provides utility functions (such as a [bulk helper](https://github.com/elastic/elasticsearch-php/issues/52#issuecomment-36906394)).  It would live outside the client itself (e.g. `\Elasticsearch\Helper\...`) and therefore be 100% opt-in.

I think this would be perfect for a semi-external helper.  Would you be willing to refactor it outside the main client and send a PR?

[...]

PR #275, opened 2015-08-07
Added helper iterators for scrolled search

arturom:
This PR adds helper Iterator classes, as discussed in issue #258.


polyfractal:
Merged, thanks!

PR #296, opened 2015-09-06
Remove monolog dependency 

xiris:
Know about a specific object logger isn't responsability of ClientBuilder.
Then this PR removes this dependency and add the type hinting.

The example below will result in fatal error:

``` php
$logger = new stdClass();
$client = \Elasticsearch\ClientBuilder::create()
    ->setLogger($logger)
    ->build();

// ... code
```

> Catchable fatal error: Argument 1 passed to Elasticsearch\ClientBuilder::setLogger() must be an instance of Psr\Log\LoggerInterface.

If we use a logger compatible with PSR-3:

``` php
$logger = new Monolog\Logger('log');
$logger->pushHandler(new Monolog\Handler\StreamHandler('logfile.log'));
$client = \Elasticsearch\ClientBuilder::create()
    ->setLogger($logger)
    ->build();

// ... code
```

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -136,7 +136,7 @@ class ClientBuilder
     }
 
     /**
-     * @param \Elasticsearch\Connections\ConnectionFactoryInterface $connectionFactory
+     * @param ConnectionFactoryInterface $connectionFactory
      * @return $this
      */
     public function setConnectionFactory(ConnectionFactoryInterface $connectionFactory)
@@ -200,10 +200,10 @@ class ClientBuilder
     }
 
     /**
-     * @param \Psr\Log\LoggerInterface $logger
+     * @param LoggerInterface $logger
      * @return $this
      */
-    public function setLogger(\Psr\Log\LoggerInterface $logger)
+    public function setLogger(LoggerInterface $logger)
     {
         $this->logger = $logger;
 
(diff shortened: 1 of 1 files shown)


src/Elasticsearch/ClientBuilder.php
```diff
@@ -221,7 +203,7 @@ public function setHandler($handler)
      * @param \Psr\Log\LoggerInterface $logger
      * @return $this
      */
-    public function setLogger($logger)
+    public function setLogger(\Psr\Log\LoggerInterface $logger)
```
rayward: The type hint can be just `LoggerInterface` as it's already imported above:

```php
use Psr\Log\LoggerInterface;
```

xiris: @rayward good, i didn't pay attention on that. thanks.

polyfractal:
Hey, sorry for the delay on this.  I was out of town for business, and then vacation.  Just now unburying my inbox. :)

Sooo...I understand the motivation for this, and agree to an extent.  But on the other-hand, the reason for the `defaultLogger` method was user-friendliness and to ease transition from the 1.0 branch (which built a logger for you implicitly).  The idea was to provide a user a method to call when they "just want to log something", without having to worry about learning and setting up Monolog on their own.

We now have a Helper namespace, perhaps we could extract it there instead and update the docs accordingly?  That way it is still available, but isn't part of the ClientBuilder itself?

xiris:
No worries, always good some vacation :D

To be honest I don't see necessity to keep the defaultLogger till in a helper.
I think we are adding "complexity" and responsibility for the SDK.

PR #312, opened 2015-10-23
1.0 Allow for custom dic to be used

### LINKED ISSUE #311: Client::SetParams() -> broken dic
Hi,

There is an undocumented configuration option for using our own dic with this client. This is great, being able to use my own dic instead of yours makes me very happy. Not that your dic is bad, I don't mean that at all. I have quite enjoyed going over it and using it for a while now, I just need more out of it is all. 

So here I am, my dic all ready to go; when I try to pass it in, it broke. Looks like we just need to allow for a new dic. 

branch: 1.0
file: src/Elasticsearch/Client.php
method: setParams();

### Current implementation 
```
        if (isset($params['dic']) !== true) {
            $dicBuilder =  new DICBuilder($hosts, $params);
        } else {
            $dicBuilder = $params['dic']($hosts, $params);
            unset($params['dic']);
        }
```

#### Suggested implementation 
```
        if (isset($params['dic']) !== true) {
            $dicBuilder =  new DICBuilder($hosts, $params);
        } else {
            $dicBuilder = new $params['dic']($hosts, $params);
            unset($params['dic']);
        }
```

dev-head:
Fixes #311

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -1562,7 +1562,7 @@ class Client
         if (isset($params['dic']) !== true) {
             $dicBuilder =  new DICBuilder($hosts, $params);
         } else {
-            $dicBuilder = $params['dic']($hosts, $params);
+            $dicBuilder = new $params['dic']($hosts, $params);
             unset($params['dic']);
         }
 


polyfractal:
Closing for now, see note in #311

PR #313, opened 2015-10-29
Fixed the bulk indexed with batches example

machour:
Was bogus and sending the batch at every loop (except when $i % 1000 was evaluated to 0, every 1000 loop)
Also changed the number of documents to index to show how to make sure the remaining docs after the loop is done are indeed indexed

### DIFF
--- a/docs/crud.asciidoc
+++ b/docs/crud.asciidoc
@@ -93,16 +93,16 @@ to batch up the requests and periodically send them:
 .Bulk indexing with batches
 [source,php]
 ----
-$params = [];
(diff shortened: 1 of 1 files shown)


polyfractal:
Doh, that's what I get for writing code blindly without testing it.  Thanks!

Would you mind signing the [Elastic CLA](https://www.elastic.co/contributor-agreement/) so I can merge this?

machour:
Done !

polyfractal:
Merged!  Thanks again :)

Issue #318, opened 2015-11-02
Fix the bulk documentation


wernight:
The last 2 examples regarding Bulk on https://www.elastic.co/guide/en/elasticsearch/client/php-api/2.0/_indexing_documents.html (crud page) seem to have completely wrong code.

I expect the code on #177 to be actually correct.

wernight:
Ok after some fiddling here is a correct piece of code:

1. Create the index
2. Run something like:

````
$params = ['body' => []];
$count = 0;

foreach ($items as $item) {
    $params['body'][] = [
        'index' => [
            '_index' => 'my-index',
            '_type' => 'my-document',
            '_id' => $item->id,
        ]
    ];
        
    $params['body'][] = [
        'title' => item->title,
        'author' => item->author,
        // ...
    ];

    if (++$count >= 1000) {
        // Send a batch of requests
        $client->bulk($params);
        $params = ['body' => []];
    }
}

if (!empty($params['body'])) {
    $client->bulk($params);
}
````

polyfractal:
Fixed in #386.

PR #319, opened 2015-11-03
Update crud.asciidoc

wernight:
Remain consistent and fix the Bulk example code for #318

### DIFF
--- a/docs/crud.asciidoc
+++ b/docs/crud.asciidoc
@@ -73,8 +73,9 @@ A simple example might look like this:
 ----
 for($i = 0; $i < 100; $i++) {
     $params['body'][] = array(
-        'index' => 'my_index',
-                    'type' => 'my_type',
+        '_index' => 'my_index',
+        '_type' => 'my_type',
+        '_id' => $i,
     );
 
     $params['body'][] = [
--- a/docs/crud.asciidoc
+++ b/docs/crud.asciidoc
@@ -72,11 +72,13 @@ A simple example might look like this:
 [source,php]
 ----
 for($i = 0; $i < 100; $i++) {
-    $params['body'][] = array(
-        '_index' => 'my_index',
-        '_type' => 'my_type',
-        '_id' => $i,
-    );
+    $params['body'][] = [
+        'index' => [
+            '_index' => 'my_index',
+            '_type' => 'my_type',
+            '_id' => $i,
+        ]
+    ];
 
     $params['body'][] = [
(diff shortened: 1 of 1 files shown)


rtuin:
This is almost correct. 

Your change will yield:
```json
{"_index": "indexname", "_type":"typename"}
```

Should be:
```json
{"index: {"_index": "indexname", "_type":"typename"}}
```

wernight:
Thanks almost missed that.

polyfractal:
Thanks!  Two things:

- Would you mind signing the [Elastic CLA](https://www.elastic.co/contributor-agreement/) so that I can merge this?

- Would you mind squashing this down to a single commit?  Not a huuuuge deal, but it'll make the history a tad cleaner.  If not, not a big deal :)

wernight:
I'm however not yet very confident it fully works. It doesn't report errors but at the moment in my case it doesn't seems to popular the database correctly. I'm still investigating; may be a follow up PR.

wernight:
Ok I found the example is correct but it should also check that `$response['errors']` is false:

````
if ($response['errors']) {
    throw new \Exception('Bulk operation failed.');
}
````

polyfractal:
I'll give it a test on my end before I merge it too, just to help verify.  

One note:  errors are not necessarily a failure.  Elasticsearch can throw a Bulk Rejection exception, which basically means the server is at capacity and it refused to index some of the docs.  They aren't really errors, but more like "try again later" messages.

A proper, sophisticated bulk helper will queue up those rejected documents in a new request, wait a random 1-5s and then retry them...repeating the process until they have all been accepted.  Then resume the bulk process.

I'm not sure we want to include that level of detail in the example though?  It's meant to show how the bulk API works, not necessarily build a 100% robust bulk method?

wernight:
I'd have in the second example (the more complete one):

````
if ($response['errors']) {
    // May be a real error, or a Bulk Rejection exception, which basically means the server
    // is at capacity and it refused to index some of the docs. Those aren't really errors,
    // but more like "try again later" messages.
    throw new \Exception('Bulk operation failed.');
}
````

polyfractal:
++ that makes sense to me.

We can link to the "Bulk Rejections" portion of the guide here too, to provide further info:  https://www.elastic.co/guide/en/elasticsearch/guide/current/_monitoring_individual_nodes.html#_threadpool_section

polyfractal:
Closing in favor of #386

PR #322, opened 2015-11-04
Update Client.php

shendongming:
if id is empty value
will drop index/type


polyfractal:
Thanks for the PR!  The required fix was actually a bit more extensive (need to check for null, empty strings, arrays of null / empty string) and needed tests.  I went ahead and implemented it myself, since its been sitting on my todo list for a while.

Thanks again, and please feel free to send more PRs in the future!

PR #334, opened 2015-11-27
json_decode(): integer overflow detected

RSol:
json_decode(): integer overflow detected

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-  "name": "elasticsearch/elasticsearch",
+  "name": "RSol/elasticsearch",
   "description": "PHP Client for Elasticsearch",
   "keywords": ["search","client", "elasticsearch"],
   "type": "library",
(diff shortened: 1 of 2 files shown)


polyfractal:
Sorry, we can't support this in ES-PHP 1.0.  The `JSON_BIGINT_AS_STRING` constant was added in PHP 5.4.0, and ES-PHP 1.0 supports down to PHP 5.3.9. :(

There are also weird problems where certain versions (or all maybe?) of Ubuntu don't have the constant since they don't agree with the license used by the underlying library.  It's complicated.  The native PHP solutions to fix this look expensive, lots of regex'ing and parsing to attempt to find numbers (but avoid numbers inside of strings).

I'm inclined to say this should be the user's responsibility, if you know your numbers might be very large.  It's simple to escape your own large numbers, but difficult/expensive to autodetect someone else's large numbers.  And ES doesn't have BigInt support either, so you should be sending them as strings anyway.

polyfractal:
Oh boy, misread the code.  This is for `json_decode()`, not encode, meaning you don't have a chance to encode bigints yourself (e.g. coming from ES).

ezimuel:
I'm closing this PR due to long inactivity (our fault, sorry for that). @RSol  I'll reopen it if you will ping me. Remember to sign the CLA, thanks!

PR #338, opened 2015-12-02
Cat shard verbosity

mariusbancos:
Added some extra verbosity for the cat shard method, like for cat recovery.


polyfractal:
++ thanks!  Don't mind the travis failure, we're having some problems with the 2.1 snapshot build at the moment.

PR #346, opened 2015-12-10
SearchHitIterator is implementing \Countable interface.

### LINKED ISSUE #341: SearchHitIterator could implement \Countable
It is quiet easy to implement `\Contable` interface in `SearchHitIterator`. In [`rewind()`](https://github.com/elastic/elasticsearch-php/blob/7b9cf8c7fcac781ccca0688fb73591f626fe1a8a/src/Elasticsearch/Helper/Iterators/SearchHitIterator.php#L55) method we already get full count of documents matching query.

I am willing to create PR for that. Any thoughts?

murnieza:
I am waiting for your opinion on calling rewind in constructor. This is needed to ensure that `count()` works before iterating over iterator.

Closes #341.

### DIFF
--- a/src/Elasticsearch/Helper/Iterators/SearchHitIterator.php
+++ b/src/Elasticsearch/Helper/Iterators/SearchHitIterator.php
@@ -14,7 +14,7 @@ use Iterator;
  * @link     http://elasticsearch.org
  * @see      Iterator
  */
-class SearchHitIterator implements Iterator {
+class SearchHitIterator implements Iterator, \Countable {
 
     /**
      * @var SearchResponseIterator
@@ -37,6 +37,11 @@ class SearchHitIterator implements Iterator {
     protected $current_hit_data;
 
     /**
+     * @var int
+     */
+    protected $count;
+
+    /**
      * Constructor
      *
      * @param SearchResponseIterator $search_responses
@@ -44,6 +49,8 @@ class SearchHitIterator implements Iterator {
     public function __construct(SearchResponseIterator $search_responses)
     {
         $this->search_responses = $search_responses;
+
+        $this->rewind();
     }
 
     /**
@@ -62,6 +69,8 @@ class SearchHitIterator implements Iterator {
         if($this->search_responses->valid() && empty($current_page['hits']['hits'])) {
             $this->search_responses->next();
         }
+
+        $this->count = $current_page['hits']['total'];
 
         $this->readPageData();
     }
@@ -135,5 +144,13 @@ class SearchHitIterator implements Iterator {
             $this->current_hit_data = null;
         }
 
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function count()
+    {
+        return $this->count;
     }
 }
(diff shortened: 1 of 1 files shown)


src/Elasticsearch/Helper/Iterators/SearchHitIterator.php
```diff
@@ -63,6 +70,8 @@ public function rewind()
             $this->search_responses->next();
         }
 
+        $this->count = $current_page['hits']['total'];
```
polyfractal: Perhaps we should put an `isset()` conditional around this to protect against unforeseen errors in the response and the `['hits']['total']` isn't set?  Just default to zero otherwise?

polyfractal:
Hmm, I don't think calling rewind is a big problem in the constructor.  It makes the iterator no longer "lazy" (e.g. calling the scroll on first iteration), but I don't think that is necessarily a problem.  We should document it however, since the scroll timeout might catch people by surprise if they instantiate the iterator but don't use it for a little while?

@arturom any opinion on this?

Left a comment about checking that `['hits']['total']` is set (in case of strange errors), but otherwise LGTM.

murnieza:
I think scroll timeout is not a problem, because when iterator will be used rewind will be called again and new window will be opened.

polyfractal:
Ah I see.  Yeah, so the only real downside then is an extra call to ES (e.g. if the query is slow, it could add some extra load).

arturom:
I agree that this Iterator should be as lazy as possible. Perhaps we can make `count()` lazy too by having it talk to the Elasticsearch cluster on its own, but only if the class hasn't done so already.

murnieza:
Moved `rewind()` call to `count()` and introduced local parameter caching to not rewind more than needed.

arturom:
hmmm.. I still don't like the fact that calling count rewinds the iterator. You could run into an issue while doing something like:
```php
foreach($my_hits_iterator as $current_index => $_hit) {
   $progress = count($my_hits_iterator) / $current_index;
}
```

It's not a very clever way to write code, but it illustrates my concern. You can of course code around this so it's not a big issue.

murnieza:
I see. So I should save count not only when count is called but on first rewind.

This should solve the problem you described.

arturom:
@murnieza ,  I was wrong, That problem I mentioned does not exist. Your condition `if ($this->count === null) { ...}` guards against that.

src/Elasticsearch/Helper/Iterators/SearchHitIterator.php
```diff
@@ -136,4 +141,24 @@ private function readPageData()
...
+        if ($this->count === null) {
+            $this->rewind();
+
+            $this->count = 0;
+
+            $current_page = $this->search_responses->current();
+
+            if (isset($current_page['hits']) && isset($current_page['hits']['total'])) {
+                $this->count = $current_page['hits']['total'];
+            }
```
arturom: Move these lines to `rewind()` so that you save the count any time you call `rewind()`, and not just when you call `count()`.

``` php
            $this->count = 0;

            if (isset($current_page['hits']) && isset($current_page['hits']['total'])) {
                $this->count = $current_page['hits']['total'];
            }
```

polyfractal:
Just fyi I'll be traveling most of the day.  I'll check in when I get home, and when @arturom gives a LGTM I'll merge :)

Thanks to both of you, really appreciate it!

arturom:
Awesome! LGTM

polyfractal:
Merged, thanks @murnieza and @arturom!

PR #351, opened 2015-12-15
Add support for generators and iterators in bulk endpoint

ramunasd:
This add support for generators and iterators in bulk endpoint. Generators can dramatically reduce memory usage and there is no need to work with big arrays.
This also improves variable usage in method bulk().

### DIFF
--- a/src/Elasticsearch/Endpoints/Bulk.php
+++ b/src/Elasticsearch/Endpoints/Bulk.php
@@ -27,25 +27,23 @@ class Bulk extends AbstractEndpoint implements BulkEndpointInterface
     }
 
     /**
-     * @param string|array $body
+     * @param string|array|\Traversable $body
      *
      * @return $this
      */
     public function setBody($body)
     {
-        if (isset($body) !== true) {
+        $this->body = '';
+
+        if (empty($body)) {
             return $this;
         }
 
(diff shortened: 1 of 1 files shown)


polyfractal:
This seems to be failing [bulk/30_big_string.yaml](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/test/bulk/30_big_string.yaml) and [bulk/40_fields.yaml](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/test/bulk/40_fields.yaml) because it doesn't handle the case where the user supplies a raw JSON string.

Otherwise, I like this.  We'll have to put this into master only for an eventual 3.0 release since 2.x supports down to PHP 5.4.  However, I want to kick 3.0 out the door a lot faster than it too to get 2.0 out (just small updates, not a complete overhaul to the core)

ramunasd:
Actually you can merge into 2.x because this improvement depends only on interface [\Traversable](http://php.net/manual/en/class.traversable.php) which is available since 5.0.0

polyfractal:
Ah nice, I overlooked that.  Heard "generators" and just assumed :)

LGTM, thanks!

PR #328, opened 2015-11-13
Update Scroll.php

amm3:
On calls to es_client->scroll(), the scroll_id would be appended as part of the URI.  Elasticsearch does not expect it here, but rather wants it in the body of the GET request.

PR #361, opened 2016-01-05
Update Scroll.php

amm3:
Reference previous request #328.

On calls to es_client->scroll(), the scroll_id would be appended as part of the URI.  This has the side effect of hitting the URL length limit (default 4kb) as scroll IDs can be in the 8kb range.  The other option is to put the scroll ID in the body of the get request, as demonstrated here.

### DIFF
--- a/src/Elasticsearch/Endpoints/Scroll.php
+++ b/src/Elasticsearch/Endpoints/Scroll.php
@@ -37,6 +37,14 @@ class Scroll extends AbstractEndpoint
         return $this;
     }
 
+    /**
+     * @return array
+     */
+    protected function getBody()
+    {
+        return $this->scroll_id;
+    }
+
     public function setClearScroll($clear)
     {
         $this->clear = $clear;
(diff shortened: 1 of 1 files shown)


polyfractal:
Thanks for reopening this (and sorry for the delay).

I looked into the test failure.  Scroll allows setting a scrollID, or setting the body explicitly.  But in this PR, the `getBody()` method always returns just the scrollID, so it fails in the [explicit body test.](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/test/scroll/10_basic.yaml#L50)

Since scroll_id isn't used in the URL anymore, I think we can:

- Remove $scroll_id variable entirely
- Set $body = $scroll_id in `setScrollID()`
- Return $body in `getBody()`

I *think* that should work, although I haven't tested it.  Mind fiddling with this a bit to get the test to pass?

amm3:
Thanks!  Admittedly, I am not well versed in reading through the travis logs.  I made your suggested changes, tested on my local environment, and we're passing now!

polyfractal:
Woo, thanks!  Merging :)

And yeah, the output from the yaml tests are... verbose and not overly user-friendly.  It's been a long evolution of tweaking the results for my own debugging purposes, which has turned it into a verbose monster :)

PR #375, opened 2016-01-18
Syntax errors

abulhol:


polyfractal:
Geez, I should really run the examples after writing docs to make sure they compile.  Thanks!

Would you mind splitting this change out into it's own branch, or squashing it down into a single commit?  Right now it also includes the autoload fix that I just merged.  When opening PRs, I like to create a dedicated branch for each PR so that it only includes the single change that I'm trying to send.

Thanks!

abulhol:
HI Zachary, please just use my changes, it is somehow easier than splitting the pull request for me. ;-)

PR #377, opened 2016-01-19
clear code

ReenExe:


polyfractal:
Some of this I like (shorthand array syntax instead of long), some of it I don't (I feel ternary conditionals are more difficult to read than the long-form version, and I like the explicitness of `isset($foo) === true`, etc).

Could you separate the changes out into individual PRs so that we can discuss them individually, instead of all jumbled up?

Also, this looks like it may just be a handful of code across the whole library?  If we change the convention for something, I would prefer it to change across the entire project.

ReenExe:
With pleasure I can walk on the project and clean code (quick and interesting).
You create a library for PHP developers and for me(us) strange to read `isset($value) === true` (isset return bool, and i see some like `bool $a = true; if ($a === true) {}` )
In this `pull request` also copies of conditions (when already know, but check again)

З задоволення можу пройтись по проекту і почистити код (шивдко та цікаво).
Ви створюєте бібліотеку для PHP розробників і для нас дивно читати повні версії замість красивого цукру.
В запиті також почистив повторні перевірки та конструкції.

polyfractal:
Gonna close this, too large and too out of date now.  Sorry.  If you want to re-send this as individual chunks we can work through it on a smaller basis.

PR #387, opened 2016-02-05
Handling children items errors

sidux:


polyfractal:
Hi there, sorry for the delay...been traveling a lot for work lately.

I'm afraid I won't be able to merge this.  The clients have a goal to remain as unopinionated as possible, and this (unfortunately) introduces some opinions that I don't think can be fixed.  For example, this would break if a random document comes back that happens to match the same structure (e.g. `status` and `items` fields).

It's also complicated when it comes to the exception itself, since a bulk may succeed (`200 OK`) but have various children responses that fail.  The request itself was still successful and shouldn't throw an exception.

Thanks for the PR though, I really appreciate it!  Perhaps we can reason a way around this, but I don't think there will be a way to implement this in an unopinionated manner.

PR #276, opened 2015-08-10
whitelist cache params

garthkerr:

PR #395, opened 2016-03-03
whitelist search per-request cache parameters

VasekPurchart:
This was already solved in #276, but I think it was then lost when merging the 2.0 branch here https://github.com/elastic/elasticsearch-php/commit/333b84898f5a96f31025022790f3b3aff222e539 .

I also added the params to the docs, so that people can find it ;)

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -890,6 +890,8 @@ class Client
      *        ['lowercase_expanded_terms'] = (boolean) Specify whether query terms should be lowercased
      *        ['preference']               = (string) Specify the node or shard the operation should be performed on (default: random)
      *        ['q']                        = (string) Query in the Lucene query string syntax
+     *        ['query_cache']              = (boolean) Enable query cache for this request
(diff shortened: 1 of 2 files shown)


polyfractal:
Sigh, I'm terrible at Git.  Good catch, thanks.  Merging, and cherry-picking back to 2.0 (which seems to be missing it somehow).

VasekPurchart:
Thanks! :) Are there any plans on a new release of `2.x`? Perhaps even a small one one with fixes, because there are quite a few since last release?

polyfractal:
Good point, I'll tag and release a version today.  Thanks for the poke...I'll ping this thread when the release is out.

polyfractal:
Small note: this was held up because of the ES 2.2 release (I wanted to make sure there aren't any test failures on 2.2 before tagging a release).  The tests are baking right now, so should have this tagged soon, barring any test problems.

VasekPurchart:
Thanks for the updates :)

polyfractal:
Just pushed and tagged as `v2.1.5`.  Thanks for your patience!

PR #407, opened 2016-04-04
Use valid SPDX license identifier

bcremer:
Use valid SPDX license identifier in composer.json to fix composer validate error.

Before:

```
$ composer validate
./composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
License "Apache 2" is not a valid SPDX license identifier, see https://spdx.org/licenses/ if you use an open license.
If the software is closed-source, you may use "proprietary" as license.
```

After:
```
$ composer validate
You are running composer with xdebug enabled. This has a major impact on runtime performance. See https://getcomposer.org/xdebug
./composer.json is valid
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
   "description": "PHP Client for Elasticsearch",
   "keywords": ["search","client", "elasticsearch"],
   "type": "library",
-  "license": "Apache 2",
+  "license": "Apache-2.0",
   "authors": [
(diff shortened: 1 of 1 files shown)


bcremer:
Signed the CLA.

polyfractal:
Oh, good catch, thanks!

PR #412, opened 2016-04-22
feat(): allow to override client instantiation

dominikzogg:
I dlike to be able to override the client itself (extending the client class for example).

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -439,7 +439,17 @@ class ClientBuilder
             };
         }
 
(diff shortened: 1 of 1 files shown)


polyfractal:
I like it!  Merging.

Out of curiosity, what are you doing with the extended class?  Is it ES-related functionality, or something unrelated?  I just ask because I'm curious if it's useful functionality the client should ship with by default :)

dominikzogg:
@polyfractal thank you for merging. We at @jobcloud already used the v1 and extended the Client to handle for example a default index. To be able using the new factory and not to have re factoring every usage, we needed a way to handle this.

PR #417, opened 2016-04-29
add reindex function

Nexucis:


src/Elasticsearch/Endpoints/Indices/Reindex.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace Elasticsearch\Endpoints\Indices;
```
polyfractal: We need to move this to the "main" namespace because the API is just `/_reindex`.  So just move this file to `src/Elasticsearch/Endpoints/Reindex.php` and modify the namespace accordingly.

src/Elasticsearch/Endpoints/Indices/Reindex.php
```diff
@@ -0,0 +1,58 @@
...
+use Elasticsearch\Endpoints\AbstractEndpoint;
+
+/**
+ * Class Analyze
```
polyfractal: Typo, this should probably be `Reindex` ?

src/Elasticsearch/Endpoints/Indices/Reindex.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+class Reindex extends AbstractEndpoint
+{
+
+    /**
+     * @return string[]
+     */
+    protected function getParamWhitelist()
+    {
+        return array();
```
polyfractal: Can you add the parameters listed here to your whitelist?  https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/reindex.json#L9

src/Elasticsearch/Namespaces/IndicesNamespace.php
```diff
@@ -275,6 +275,25 @@ public function snapshotIndex($params = array())
     }
 
     /**
+     * @param $params array Associative array of parameters
```
polyfractal: Similarly to the endpoint, we need to move this into `Client.php`, since it is part of the "main" namespace.

polyfractal: Oh, can we also add the the list of parameters here too?  You can copy the formatting from another of the methods above

polyfractal:
Oh, cool, I assumed your PR would be for a "reindex helper" using scan/scroll, but this is for the Reindex API.  Thanks!

I left a few comments, but it's all just logistical stuff.  We need to move this to the "main" namespace because the api is `/_reindex`, so it doesn't live under the indices namespace.  And a few tweaks to the whitelist / comments.

Otherwise it LGTM!  Thanks!

Nexucis:
thanks a lot for the code review !

I hope it should be alright now . By the way, I also developped a reindex helper according to [this link]( https://www.elastic.co/blog/changing-mapping-with-zero-downtime#_reindexing_your_data). But I think I still have work to do before I can propose to you the code. Maybe for another PR :) ?

polyfractal:
LGTM, thanks!  Merging

> By the way, I also developped a reindex helper according to this link. But I think I still have work to do before I can propose to you the code. Maybe for another PR :) ?

++ a separate PR would be perfect. :)

PR #418, opened 2016-05-14
add index helper

Nexucis:
Hello,

As i said in a previous PR, i'm sharing an index helper which provide some useful method to provide an index usage without downtime  (i think, may be you have not the same opinion :) ).

I hope you will like it.

### DIFF
--- a/src/Elasticsearch/Helper/Nodowntime/Exceptions/IndexNotFoundException.php
+++ b/src/Elasticsearch/Helper/Nodowntime/Exceptions/IndexNotFoundException.php
@@ -1,0 +1,19 @@
+<?php
+namespace Elasticsearch\Helper\Nodowntime\Exceptions;
+
+
+use Elasticsearch\Common\Exceptions\ElasticsearchException;
+
+/**
+ * IndexNotFoundException
+ *
+ * @category Elasticsearch
+ * @package  Elasticsearch\Helper\Nodowntime\Exceptions
+ * @author   Augustin Husson <husson.augustin@gmail.com>
+ * @license  http://www.apache.org/licenses/LICENSE-2.0 Apache2
+ * @link     http://elastic.co
+ */
+class IndexNotFoundException extends \Exception implements ElasticsearchException
+{
+
+}
--- a/src/Elasticsearch/Helper/Nodowntime/Exceptions/IndexNotFoundException.php
+++ b/src/Elasticsearch/Helper/Nodowntime/Exceptions/IndexNotFoundException.php
@@ -5,7 +5,7 @@ namespace Elasticsearch\Helper\Nodowntime\Exceptions;
 use Elasticsearch\Common\Exceptions\ElasticsearchException;
 
 /**
- * IndexNotFoundException
+ * IndexNotFoundException thrown when an index is not found
  *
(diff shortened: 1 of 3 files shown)


Nexucis:
Hi, sorry for these commits, , i had to fix phpdocs and improve the code. I think it should be better now. 
I hope you don't take this wrong.

polyfractal:
Sorry for the delay, been a busy week :)  I'll try to take a look at this today, but it'll probably slide to next week.

I skimmed it, and I'm curious why you've implemented methods for things like updating/deleting docs, searching, etc?  It looks like they are just pass-through methods for the underlying update/delete/etc methods on the client?

Nexucis:
Thanks for the reply :). I appreciate a lot, thanks ! 
You have right, these methods (like update/delete docs) are just pass-through methods. In my mind, i was thinking that it could be great if you have a complete helper which allows you to manipulate your index via an alias. And may be it could be easier to use these methods than directly using the API. 
Finally, it can provide a complete and simple example for how using elasticsearch with no downtime.
 
But may be, it's not as useful than i was thinking. What do you think ?

Nexucis:
Hello :). 

I'm sorry for the inconvenience, but I would like to know if your are still interest by this PR ? I know you are busy, that's why I hope you don't mind answering this question !

polyfractal:
Hi, sorry!  Been super busy.  I'll try to take a closer look at this ASAP.

Nexucis:
I also dust this PR :). Can you give me a feed back when you have time? 

Thanks in advance!

polyfractal:
Hi.  Apologies again for the delay... big PRs are hard to find time to look at.

I've looked through the PR and I don't think it is a good fit for ES-PHP.  It looks like the code itself is reasonable... the main issues are the size and the "opinionated" nature.

The client tries to stay relatively unopinionated; it provides a raw view of the Elasticsearch APIs without a lot of opinions about how the APIs should be used.  Helpers like the scroll helper are acceptable because they are just sugar over what a developer would write themselves.  This index helper, while useful, puts a lot of opinions about how indices/aliases should be used, what conventions are held to, etc.

The other main reason is simply size: it's a lot of code that this project would become responsible for maintaining.  There aren't any tests to verify that it works, and I won't personally be using the helper to verify it either.  That makes the code fairly unmaintainable by me or other contributors.

So for those two reasons I don't think it will be appropriate to merge this PR. Sorry for the long delay, that was definitely my fault for not getting back to you sooner.

BUT, with that said, I think this is useful code for people to use.  Why not open a small repository of your own that hosts the helper?  It can depend on the ES-PHP library and other people can install it alongside their code if they want to use it. We can add a link in the "Community" section of the ES-PHP documentation to help people find the helper.  Thoughts?

Nexucis:
Thanks a lot for your response! It was really instructive :) 

It sounds good to me and I'll do it asap (I'm currently in a vacation and unfortunately I can't do it right now) 

Once it's done, can I post on the relative issue? Or do you want a simply email with the link to the repository?

polyfractal:
Enjoy your vacation!

When it's done, feel free to either open an issue with the link, or file a PR modifying the documentation page with a link.  Either works :)

PR #420, opened 2016-05-26
Issue-419: Support of 2.3

Tezd:
Support of 2.3 Elasticsearch database for elasticsearch-php library


karmi:
Hi @Tezd, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-php/pull/420.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Tezd:
Hi, sure I am on top of this :) Was updating related issue.

polyfractal:
Woah, this is big!  I'll start looking through this, thanks!

Is this the result of running ParseSpec?  To be honest, I haven't used that tool since the creation of the client when the boilerplate was generated, I'm kinda amazed it still (sorta) works.  I usually just update the changes by hand (usually adding a few whitelist parameters, or new endpoint)...which is why a lot of those param comments are outdated I suppose :)

Although now that I think about it, we need to update the whitelist's themselves, to match the comments.  I'm guessing the ParseSpec utility didn't/doesn't do that.  I can start working on that.

Regarding the removed endpoints: we'll have to leave them in-place, unfortunately.  When the 2.x branch was created, the branching / compatibility methodology wasn't well defined.  So the 2.x branch technically supports down to ES 1.x.  If we remove those we may break someone's code.

Also it would require a major version bump of ES-PHP, but we're now tied to ES release cycle and can only bump major when they do.

Let's just label them with `@deprecated` and rip them out for 5.x?

Thanks so much for this!  Updating the generator in particular is great, I should really be running that on a regular basis :)

Tezd:
Understood i will put them back in tomorrow. Yes, new endpoints result of ParseSpec :) Btw I already updated whitelists :)

Tezd:
Do I also need to bring back old parameters for whitelist?

antonbormotov:
@polyfractal, is there a reason to keep unsupported APIs endpoints and mark them as deprecated? Even doing so, will not make branch 2.0 back-compatible with Elastic Search 1.x.

polyfractal:
> Understood i will put them back in tomorrow. Yes, new endpoints result of ParseSpec :) Btw I already updated whitelists :)

Great, thanks!  No rush, I'm off today, so I won't have a chance to test this until monday.

> Do I also need to bring back old parameters for whitelist? 

Yep, otherwise someone relying on the client may get an unexpected error if they upgrade minor versions

> @polyfractal, is there a reason to keep unsupported APIs endpoints and mark them as deprecated? Even doing so, will not make branch 2.0 back-compatible with Elastic Search 1.x.

It's to avoid bwc breaks at the ES-PHP level.  Elasticsearch 1.x and 2.x are largely compatible at the REST api level ([very few breaking changes](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-2.0.html)).  E.g. all the URL endpoints are the same, parameters are mostly the same, etc.  There are *new* endpoints in 2.x which won't work in 1.x obviously, but the client itself is backwards compatible and can be used down to 1.x as long as your code only uses the feature set supported by the server.  And since the client doesn't validate the body, none of the larger changes (like the mapping changes) actually affected the client.

Note: This is how minor versions work too: new functionality is added as ES adds it over time.  If you are on a slightly older version of Elasticsearch, but the most recent ES-PHP, you may not be able to use the full set of features until you upgrade.

What I should have done is made it clear that ES-PHP 2.x is only to be used with ES 2.x.  But at the time we weren't syncing releases, and since ES-PHP 2.x was backwards compatible, I figured it would be better to support a wider range.

But now we're in a situation where we cannot remove endpoints for fear of breaking semantic versioning.  If we remove an endpoint, we have to bump the major version since it is a change in functionality and may break code.  But releases are now tied to ES releases*, so I cannot release until ES does.

So at the end of the day, it's basically because I didn't plan out the version matrix well, and we're stuck supporting down a version of ES.  It's not a great situation, but it is what it is.  Oops :(

* Technically, we could release an ES-PHP 3.0 to include these changes, but ES-PHP 5.0 will be released shortly to coincide with the ES 5.0 ...  so I'm not sure I see much utility in it.

SergeyShcherbin:
@polyfractal updated branch to be back compatible

polyfractal:
Looked through it this morning, ran the tests, looks good!  Merging!

I'll re-generate the docs, and then cherry-pick some of these changes to master so we can re-generate the params/whitelists/etc over there too.  Thanks!

polyfractal:
Hm, so I applied the new ParseSpec to master without really looking at it closely, and realized afterwards that it still has most of the problems that made me stop using it to begin with (no camel-casing, no special cases for specialized abstract endpoints, etc).  I had to back the changes out because it broke too much.

These are obviously fixable, but I noticed that your patch didn't need this fixing ... did you do something different with the output?  Similarly, ParseSpec doesn't update the namespace comments (I'd forgotten about that) ... did you write a script to generate them?

If yes to either of those questions, I'd love to incorporate them :)

SergeyShcherbin:
@polyfractal I just fixed path from where it took api specs since it was pointing to a wrong folder. Yes, I have a script which generates phpdoc block I can incorporate it into ParseSpec file. Will do it on weekends.

polyfractal:
That'd be great, thanks!  No rush, but it'd be a useful tool.  I'm going to keep working on the template to handle the edge cases, I'd like to get it to the point where we can just run the tool to update between versions unless something major has changed.

PR #422, opened 2016-06-02
Make it easier to note that 'client' is an array of options.

stavarengo:


stavarengo:
To me it appears as if had not signed the CLA, but I already did.
Is this some kind of a delay?

polyfractal:
Sometimes it get's wonky (Adobe's EchoSign API is pretty much the definition of what not to do with APIs).  I just checked manually and you're good to go. :)

Thanks for the PR!

PR #426, opened 2016-06-28
Fixed snapshot uri

cyrildev:

### DIFF
--- a/src/Elasticsearch/Endpoints/Snapshot/Status.php
+++ b/src/Elasticsearch/Endpoints/Snapshot/Status.php
@@ -60,7 +60,7 @@ class Status extends AbstractEndpoint
      */
     protected function getURI()
     {
-        if (isset($this->snapshot) === true && isset($this->repository) !== true) {
+        if (isset($this->snapshot) && !isset($this->repository)) {
             throw new Exceptions\RuntimeException(
                 'Repository param must be provided if snapshot param is set'
             );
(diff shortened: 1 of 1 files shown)


src/Elasticsearch/Endpoints/Snapshot/Status.php
```diff
@@ -60,7 +60,7 @@ public function setSnapshot($snapshot)
      */
     protected function getURI()
     {
-        if (isset($this->snapshot) === true && isset($this->repository) !== true) {
+        if (isset($this->snapshot) && !isset($this->repository)) {
```
polyfractal: Would you mind reverting this change?  The library follows the `foo === true` convention everywhere, and avoids using the inline shorthands (implicit true, `!` negation).

src/Elasticsearch/Endpoints/Snapshot/Status.php
```diff
@@ -70,11 +70,12 @@ protected function getURI()
         $snapshot   = $this->snapshot;
         $uri        = "/_snapshot/_status";
 
-        if (isset($repository) === true) {
-            $uri = "/_snapshot/$repository/_status";
-        } elseif (isset($repository) === true && isset($snapshot) === true) {
+        if (isset($repository, $snapshot)) {
```
polyfractal: Same here:  `isset($repository, $snapshot) === true`

And below too vvvvvv

polyfractal:
Thanks!  Would you mind [Signing the CLA](https://www.elastic.co/contributor-agreement/) so that we can merge this?

Left a quick comment about formatting too.

polyfractal:
Actually, upon closer look, this appears identical to the current code?  Closing, feel free to reopen if I'm missing something :)

PR #430, opened 2016-07-11
Fixes #398: Switch method to POST if request has a body

### LINKED ISSUE #398: Search Method Changing
Is it possible with the current configuration to change the default search method from GET to POST? It looks as though it is hard coded here:
https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Endpoints/Search.php#L103

I ran into a scenario where I had different search results return for the same query depending on the method. It appears to default to GET. When I change it to POST I get the expected results. Thanks.

lifeofguenter:
This PR will set the http-method to `POST` on search queries, if a data-body has been supplied (else it will default to `GET`).

(also adding ZendStudio files to the `.gitignore`)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ composer.phar
 
 #editor related
 .idea
+/.settings/
+/.buildpath
+/.project
 
 # OS generated files
 .DS_Store
--- a/src/Elasticsearch/Endpoints/Search.php
+++ b/src/Elasticsearch/Endpoints/Search.php
@@ -16,6 +16,8 @@ use Elasticsearch\Common\Exceptions;
  */
 class Search extends AbstractEndpoint
 {
+    protected $method = 'GET';
+
     /**
      * @param array $body
      *
@@ -29,6 +31,7 @@ class Search extends AbstractEndpoint
         }
 
         $this->body = $body;
+        $this->method = 'POST';
 
         return $this;
     }
@@ -102,6 +105,6 @@ class Search extends AbstractEndpoint
      */
     public function getMethod()
     {
-        return 'GET';
+        return $this->method;
     }
 }


polyfractal:
Sorry for the extreme delay in responding.  I don't think this is necessary, as ES supports GET + Body.  And if we were to support a switch like this, it should apply universally to all requests that support GET + Body too, not jus search.

Closing, thanks though!

kamerynn:
If changes were made to support a configuration change to use POST instead of GET for these types of requests (rather than a change to the Search endpoint specifically, without any configuration option) would you consider accepting it?

We've run into the same issues when placing Squid proxy in front of the requests. I agree that the library should stay close to what Elasticsearch accepts, but from our digging, there really is no good solution around this problem.

polyfractal:
Would you be ok with a per-request config option?  E.g. something like:

```php
$params = [
    'index'  => 'test_missing',
    'type'   => 'test',
    'id'     => 1,
    'client' => [ 'method' =>'POST' ]  //<-- here
];
$client->search($params);
```

There's already plenty of precedent for per-request config options, so this would fit in just fine.  A little more hassle since you're app would have to set it each time, but not toooo onerous I think?

If that works for you, I could knock that out easily/quickly. Or you can send a PR if you're wanting to contribute. :)

(Technically the method can be overridden by setting the appropriate `CURLOPT_POST` curl option, but that's pretty gross.  I'd be fine with providing a simpler way to override it)

kamerynn:
Thanks for responding so quickly!

I had tried setting `CURLOPT_CUSTOMREQUEST => 'POST'` during my testing, which worked for http, but not https, oddly enough. Both http(s) were being proxied, but it was getting blocked by the proxy (the 411 error that was mentioned) for https.

When I changed the string in the `getMethod` method, like the commit in this PR, it worked for both.

I will do some additional testing tonight to see if `CURLOPT_POST` acts any different! I figured there might be some discrepancy between setting `CURLOPT_CUSTOMREQUEST` vs. setting that string to `'POST'`. I will let you know - if `CURLOPT_POST` works, and this change would provide a cleaner way of setting the same thing, that would be great!

kamerynn:
We no longer need to proxy thru to ES, but I'd like to take a crack at contributing this change! I will work on it this weekend. Thanks for your help

polyfractal:
Hey sorry for the delay, been a zoo around here in the leadup to Elasticon.  

++ a PR for this change, it'd be very helpful!  Lemme know if you have any questions about implementation.  I imagine it'll end up similar to [custom headers](https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Connections/Connection.php#L118), where you can just check if the "method" param is set on a per-request basis and adjust appropriately.

I'll be relatively inactive for the next week due to Elasticon, just as an FYI if I go radio-silent :)

PR #443, opened 2016-07-28
Decouple endpoint from transport

joelwurtz:
This PR replace #383 and is a first step for #350 

I decided to do small steps as #383 was too big of a change (there will be others PR if this one is merged, to use PSR7 and other stuffs), and also this one has less "politics"

Purpose of this change it's to decouple entirely transport from the endpoint, so endpoint will never have to know about a transport.

The transport, now, has the responsability to transform an endpoint to a request.

Doing this will allow for better reusability of the endpoint (i.e. create a PSR7 request from it).

This change is BC and add new method to the endpoint contract (getUri, getMethod, getParams, getBody are now public method instead of protected / only propriety)

For the BC part this may be possible to have a deprecation layer, however the maintenance will be harder and IMO es 5.0 will certainly have BC so this might be the right time to do this ?

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -89,9 +89,9 @@ class Client
 
         /** @var \Elasticsearch\Endpoints\Info $endpoint */
         $endpoint = $endpointBuilder('Info');
-        $response = $endpoint->setParams($params)->performRequest();
+        $endpoint->setParams($params);
 
-        return $endpoint->resultOrFuture($response);
+        return $this->transport->performEndpoint($endpoint);
     }
 
     /**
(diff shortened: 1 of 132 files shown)


joelwurtz:
Okay nevermind, should have update my fork xD

polyfractal:
:D  Sorry, I should have updated those tickets, entirely forgot!  I needed the refactoring for a separate issue (to make injecting new namespace modules a whole lot easier), so it seemed best to just knock it out while I was deep in the code.

It's not a 100% clean situation yet (e.g. resultOrFuture() still lives in the endpoint, which is a bit odd placement considering the refactor).  But it's a start :)

Thanks for the PR though!  Hope you didn't spend too much time on it

PR #444, opened 2016-07-29
Only show log when test fails

joelwurtz:
This is mainly to show full log of elasticsearch integration test only when a test fail, also this better track which integration test fail with the source code (as it show the output just above the exception)


polyfractal:
I like this :)  Merging!  

Also, apologies that you had to slog through this horrible, terrible, no-good piece of code.  It's truly the stuff of nightmares.  Alas, it works (mostly) and it's not a big enough problem to rewrite :/

joelwurtz:
> Also, apologies that you had to slog through this horrible, terrible, no-good piece of code. It's truly the stuff of nightmares. Alas, it works (mostly) and it's not a big enough problem to rewrite :/

No worries, sure it's not perfect, but it's far from being horrible :)

PR #445, opened 2016-07-29
Move resultOrFuture from endpoint to transport

joelwurtz:
This remove resultOrFuture from the endpoint, transport deal with that, so now an endpoint does not have anymore logic on the request / response only config parameters (options).


polyfractal:
Thanks!  I'm slogged with other work today, but I'll look at this (and the other PRs) tomorrow.  :)

polyfractal:
LGTM! Merging

PR #447, opened 2016-07-31
Split, refactor and fix some tests

joelwurtz:
Hello,

I was curious how integration test was working so i rewrite the test class, IMO it is more readbable and debuggable (but that's subjective) , also i start from scratch by looking as few as possible as the existing one.

I also split into 2 phpunit files, now by default integrations tests are not run, but you can use the phpunit-integration.xml file to run integrations tests (i add it to travis).

I also try to fix tests / code, but from what i have seen, most of the remaining tests that are failing are yaml files not updated with the current contract of elasticsearch.

Also there is no more run in async as, IMO, it does not add value to do that in integration.

### DIFF
--- a/src/Elasticsearch/Endpoints/Ingest/Pipeline/Get.php
+++ b/src/Elasticsearch/Endpoints/Ingest/Pipeline/Get.php
@@ -23,7 +23,9 @@ class Get extends AbstractEndpoint
     public function getURI()
     {
         if (isset($this->id) !== true) {
-            return '/_ingest/pipeline/*';
+            throw new Exceptions\RuntimeException(
+                'id is required for GetPipeline'
+            );
         }
 
         $id = $this->id;
--- a/src/Elasticsearch/Endpoints/Ingest/Pipeline/Get.php
+++ b/src/Elasticsearch/Endpoints/Ingest/Pipeline/Get.php
@@ -23,10 +23,9 @@ class Get extends AbstractEndpoint
     public function getURI()
     {
         if (isset($this->id) !== true) {
-            throw new Exceptions\RuntimeException(
-                'id is required for GetPipeline'
-            );
+            return '/_ingest/pipeline/*';
         }
+
         $id = $this->id;
         $uri = "/_ingest/pipeline/$id";
 
--- a/tests/Elasticsearch/Tests/YamlRunnerTest.php
+++ b/tests/Elasticsearch/Tests/YamlRunnerTest.php
@@ -26,7 +26,7 @@ use Symfony\Component\Yaml\Yaml;
  * @license    http://www.apache.org/licenses/LICENSE-2.0 Apache2
  * @link       http://elasticsearch.org
  */
-class YamlRunnerTest2 extends \PHPUnit_Framework_TestCase
+class YamlRunnerTest extends \PHPUnit_Framework_TestCase
 {
     /** @var  Parser */
     private $yaml;
--- a/tests/Elasticsearch/Tests/YamlRunnerTest.php
+++ b/tests/Elasticsearch/Tests/YamlRunnerTest.php
@@ -10,6 +10,7 @@ use Elasticsearch\Common\Exceptions\Forbidden403Exception;
 use Elasticsearch\Common\Exceptions\Missing404Exception;
 use Elasticsearch\Common\Exceptions\RequestTimeout408Exception;
 use Elasticsearch\Common\Exceptions\ServerErrorResponseException;
+use GuzzleHttp\Ring\Future\FutureArrayInterface;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Finder\SplFileInfo;
 use Symfony\Component\Yaml\Exception\ParseException;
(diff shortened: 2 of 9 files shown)


joelwurtz:
Also now tests are marked skip or incomplete given the following behavior : 

 * Test is skip only with the operation "skip" matching
 * Test is incomplete if yaml could not be parsed or if the method does not exists in the lib

src/Elasticsearch/Endpoints/Ingest/Pipeline/Get.php
```diff
@@ -23,10 +23,9 @@ class Get extends AbstractEndpoint
     public function getURI()
     {
         if (isset($this->id) !== true) {
-            throw new Exceptions\RuntimeException(
```
polyfractal: Why the removal?  ID is required for the GetPipeline API: https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/ingest.get_pipeline.json#L12

tests/Elasticsearch/Tests/YamlRunnerTest.php
```diff
@@ -77,727 +67,621 @@ public static function setUpBeforeClass()
...
+     *
+     * @param $operation
+     * @param $lastOperationResult
+     * @param $testName
+     */
+    public function operationSkip($operation, $lastOperationResult, $testName)
     {
-        $ret = array();
+        if (property_exists($operation, 'features')) {
+            // No features supported
```
polyfractal: This is likely my fault, as the old method called it a "whitelist", when in reality it was a blacklist.  E.g. a list of features *not* supported.  

We should setup a proper whitelist so that the runner knows what features are supported.  It should support the `stash_in_path` feature to be compatible with the old runner, and soon i'll need to add support for headers and warnings

tests/Elasticsearch/Tests/YamlRunnerTest.php
```diff
@@ -77,727 +67,621 @@ public static function setUpBeforeClass()
...
-     * word in PHP. :/
+     * Format a regex for PHP
      *
-     * @param $file
-     * @return mixed
+     * @param $regex
+     *
+     * @return string
      */
-    private function checkForList($document)
```
polyfractal: This `checkForList()` method will need to be replicated in the new runner, somehow.  The `tasks.list/10_basic.yaml` invokes the [Tasks.List API](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/tasks.list.json), which isn't compatible with PHP (since the method call would be `$client->tasks()->list([...])`, and `list()` is a reserved PHP function).  

The method is renamed `get()` in the client, and we have to do the substitution before calling.

polyfractal:
I like it!  Much cleaner :)

I think we will likely want to put back the manual skip-list.  The two manually skipped tests were put there, if I remember correctly, because of differences in regex engine behavior.  I decided it was easier to just skip the tests than try to work around the regex differences.  These tests are run by Java, Perl, Python, Ruby, JS, Groovy and PHP... regex differences are almost guaranteed :)

I'll investigate closer when I get a chance, it may be an easy fix, or something has changed recently.

There are some yaml parsing problems, which may have been special-cased before as well.  I'll investigate when I get a chance here too.

> Also there is no more run in async as, IMO, it does not add value to do that in integration.

I'm not entirely convinced about this to be honest.  If it's not tested, it's not really supported.  I originally added it because mocking out the behavior of libcurl is non-trivial (or impossible), and weird things pop up all the time with curl "in the wild" so to speak.

Otherwise this looks really great, and we can iterate on the little problems that pop up over time.  Thanks!

src/Elasticsearch/Endpoints/Ingest/Pipeline/Get.php
joelwurtz: It was to make this test working : https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/test/ingest/10_basic.yaml#L48

But maybe the test is wrong ?

tests/Elasticsearch/Tests/YamlRunnerTest.php
joelwurtz: Maybe we can do a mapping config for this (with a simple array) ? So if there is new cases of specific keywords it easy to add ?

tests/Elasticsearch/Tests/YamlRunnerTest.php
joelwurtz: So i should not skip if the feature is `stash_in_path` (will do a simple array on top of the class)

src/Elasticsearch/Endpoints/Ingest/Pipeline/Get.php
polyfractal: Emmm... yeah that's strange.  I'll inquire :)

Either the test is wrong, or more likely, that's a new "get all pipelines" endpoint (`GET /_ingest`) which isn't documented, and so the REST Spec is wrong :)

Let's leave it alone for now while I poke some people :)

tests/Elasticsearch/Tests/YamlRunnerTest.php
polyfractal: ++  yep, let's do a simple whitelist array at the top that contains `stash_in_path`, and we can add more when features are added.

There's only one occurrence of `stash_in_path` ([nodes.info](https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/test/nodes.info/20_transport.yaml#L16)), but the functionality is used all over the place in the form of `is_true: nodes.$master.roles`.  So it shouldn't really break anything.

tests/Elasticsearch/Tests/YamlRunnerTest.php
polyfractal: ++

joelwurtz:
> I'm not entirely convinced about this to be honest. If it's not tested, it's not really supported. I originally added it because mocking out the behavior of libcurl is non-trivial (or impossible), and weird things pop up all the time with curl "in the wild" so to speak.

It's mainly because right now wait is called each time : https://github.com/elastic/elasticsearch-php/blob/master/tests/Elasticsearch/Tests/YamlRunnerTest.php#L655

So we don't really test the integration with elasticsearch here, but more the underlying library. Maybe we should add a single integration test on async, and not running the whole test suite (as it double the time of the integration suite)

AFAIK for curl mocking i know these guys are doing it : https://github.com/php-vcr/php-vcr but it's not so clean (they replace the code before loading the php file, so `curl_exec` would become something like  `VCR::curl_exec`) I don't think we should do that xD

src/Elasticsearch/Endpoints/Ingest/Pipeline/Get.php
joelwurtz: I revert this and will do an other PR for this (so it doesn't block this PR)

joelwurtz:
I readd async tests

polyfractal:
Fair points about async.  Perhaps we just hide it behind a flag so that it doesn't double the test time for an average contributor?

I think this looks great!  I need to wrap up something else, then I'll merge.  Thanks so much for tackling this! :)

PR #450, opened 2016-08-03
Tests cleaned up a little

carlcasbolt:
Removed the empty `foreach()` loops and replaced assertEquals on the array index with an `assertCount()`


polyfractal:
++

PR #454, opened 2016-08-16
Update class Client.php call for case sensitive

txandy:
Update call method for case sensitive systems in ReIndex method.

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -708,8 +708,8 @@ class Client
 
         /** @var callback $endpointBuilder */
         $endpointBuilder = $this->endpoints;
-        /** @var \Elasticsearch\Endpoints\Reindex $endpoint */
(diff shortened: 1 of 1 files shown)


polyfractal:
The endpoint is `Reindex` with a lowercase 'i' though?  

https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Endpoints/Reindex.php#L14

txandy:
Yes my bad. its solved in master branch, i use v2.2.1

dalder:
true, seems to be solved in master. could this be fixed in branch 2.x too?

polyfractal:
Agreed, we should backport this (sorry for delay, this slid through my inbox while traveling and I didn't notice)

Busy today, but tomorrow is my "PHP client day", so I'll backport then unless someone beats me to it with a PR :)

dcarbone:
👍  for getting this into a 2.x release :)

PR #456, opened 2016-08-25
Allow ConnectionFactory to be override

tchiotludo:
Changing the ConnectionFactory with the `ClientBuilder::setConnectionFactory` can't work since in AbstractConnectionPool class, it was not the `ConnectionFactoryInterface` that is waiting but the `ConnectionFactory`

Also added 2 missing properties declaration.


polyfractal:
Oof, you're right, thanks for the PR.  Merging!

tchiotludo:
Thanks for the merge! 
Just a question, do you want also a PR for the 2.x version ? or the dev-master will be compatible with ES 2.X ?

polyfractal:
I'm on the run at the moment and was going to cherry-pick the merge to 2.x when I got to my hotel tonight, but if you want to send a PR against 2.x branch that'd be even better! :)

dev-master (soon to be 5.x) should theoretically *mostly* work with 2.x, but there are a few backwards incompatible things (mostly deprecated/removed APIs), so it'd be good to merge this change to 2.x directly.

PR #466, opened 2016-09-16
Bug/reindex casing

dcarbone:
Small bugfix to allow the `Reindex` endpoint to be used in a case-sensitive environment.

On a side-note, was not super sure which branch to create the PR against, I assumed `2.0` but let me know if this is the incorrect one!

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -1582,12 +1582,12 @@ class Client
      *
      * @return array
      */
-    public function reIndex($params = [])
+    public function reindex($params = [])
     {
         /** @var callback $endpointBuilder */
         $endpointBuilder = $this->endpoints;
 
-        /** @var \Elasticsearch\Endpoints\ReIndex $endpoint */
+        /** @var \Elasticsearch\Endpoints\Reindex $endpoint */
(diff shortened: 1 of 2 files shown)


karmi:
Hi @dcarbone, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-php/pull/466.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

dcarbone:
Ah!  Oops.  Accidentally signed with my work email.  Both have been added!

polyfractal:
Thanks!  I thought I had fixed this in 7ca34ab496d51aeb5058c6704f34c81516b5f681, but it looks like I only changed the method call and not the file name.

Would you mind editing the PR so that it just includes your commit, and not the two unrelated commits?

2.0 branch is perfect, this has actually been fixed in master I just forgot to backport it to 2.0 :)

dcarbone:
Sigh, I'm just doing all the scrubby things today!  PR modified :)

polyfractal:
No worries, I'm still waking up myself  ☕  

Merging, thanks!

Issue #469, opened 2016-09-23
Docs: incorrect path to Namespaces/Indices.php


joostvanderborg:
### Summary of problem or feature request

The docs seem to refer to Namespaces/Indices.php, which should be Namespaces/IndicesNamespace.php . I'd be happy to edit myself and submit a pull-request, but the contributor guidelines say to submit an issue, so here it is.

### System details

Not relevant
- ES 2.4 (ES-PHP client 2.0)

polyfractal:
Whoops!  Go ahead and send a PR, I'll merge when I get a chance (on the go atm).

Also ++ for reading the contributor guidelines, so few do! :)

polyfractal:
Closing since the PR was merged.  Thanks :)

PR #470, opened 2016-09-23
Update index-operations.asciidoc

joostvanderborg:
The path to the file indices.php seems incorrect, should be IndicesNamespace.php. (See #469)


polyfractal:
Thanks!

PR #473, opened 2016-09-27
fix typo

bhavin-nakrani:

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -7,6 +7,7 @@ use Elasticsearch\Common\Exceptions\InvalidArgumentException;
 use Elasticsearch\Common\Exceptions\Missing404Exception;
 use Elasticsearch\Common\Exceptions\TransportException;
 use Elasticsearch\Endpoints\AbstractEndpoint;
+use Elasticsearch\Namespaces\AbstractNamespace;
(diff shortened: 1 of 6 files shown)


polyfractal:
Hm, I think argument comments are only useful if there are descriptions attached to them. The IDE can tell me what the arguments are, so adding comments only becomes helpful if the comments describe the args themselves.

src/Elasticsearch/Client.php
```diff
@@ -1311,6 +1312,10 @@ public function extractArgument(&$params, $arg)
         }
     }
 
+    /**
+     * @param $name
+     * @param $var
+     */
     private function verifyNotNullOrEmpty($name, $var)
     {
         if ($var === null) {
```
vrkansagara: what if we use ` if(strcmp($var, null) === 0 ){ ` stric comparision here.

Oliboy50: I don't understand what's the problem here.

Besides: 
> If you rely on strcmp for safe string comparisons, both parameters must be strings, the result is otherwise extremely unpredictable.
For instance you may get an unexpected 0, or return values of NULL, -2, 2, 3 and -3.

(from the top comment on http://php.net/manual/fr/function.strcmp.php)

ezimuel:
Closing due to inactivity.

PR #474, opened 2016-10-01
Added support for PHP 7.1

grogy:
PHP 7.1 [comming soon](http://php.net/archive/2016.php#id2016-09-29-1), so this library can be compatible :-)


polyfractal:
Indeed!

PR #481, opened 2016-10-25
add JSON_PRESERVE_ZERO_FRACTION for Json_encode

moknomo:
Problem discuss here
http://stackoverflow.com/questions/30726049/php-how-to-prevent-json-encode-rounding-number-automatically
This is important with dynamic mapping, when the first stored value is 0.0 but the field should be fractional. In current version (float)0.0 converted to (int)0 and elastic created field with long type.

Warning! JSON_PRESERVE_ZERO_FRACTION option available since PHP 5.6.6.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     }
   ],
   "require": {
-    "php": "^5.6|^7.0",
+    "php": "^5.6.6|^7.0",
     "psr/log": "~1.0",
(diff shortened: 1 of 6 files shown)


polyfractal:
++ I've wanted to enable this in the past, but version requirements were problems with earlier versions.  I'm fine with bumping master/5.x's min version to 5.6.6 in order to get this in.

Thanks!  Merging

PR #482, opened 2016-10-25
Encode Url to accept UTF-8 Char

Nexucis:
Hello,

I'm currently working with alias managing. And when a utf-8 char (like japan char : 金魚) is used for creating alias or search or anything in relation with alias, there is an error (like alias notfound). So I fixed it by putting the method `urlencode` in the method `setName` in the Endpoint Alias (for example : in the class `Elasticsearch\Endpoints\Indices\Alias\Get`) like this :

``` php
$this->name = urlencode($name);
```

Aaand, juste after fixing this bug, I saw the class `AbstractAliasEndpoint` where I find my fix. So I remove my fixe and extend all EndPoint Alias to the AbstractAliasEndpoint :

``` php
class Exists extends AbstractAliasEndpoint{
}
```

After that, I thought that maybe I could apply the same method in other endpoint. And that's why, you have another Pull Request with a lot of changes :). 

I hope you will enjoy it.

PS : I'm sorry to do not do some unit commits but one big commit (I have some trouble with the command `git --amend`. If you don't like it, I could do it again with unit commit.

### DIFF
--- a/src/Elasticsearch/Endpoints/Cat/Aliases.php
+++ b/src/Elasticsearch/Endpoints/Cat/Aliases.php
@@ -29,7 +29,7 @@ class Aliases extends AbstractEndpoint
             return $this;
         }
 
-        $this->name = $name;
+        $this->name = urlencode($name);
 
         return $this;
     }
--- a/src/Elasticsearch/Endpoints/Cat/Allocation.php
+++ b/src/Elasticsearch/Endpoints/Cat/Allocation.php
@@ -29,7 +29,7 @@ class Allocation extends AbstractEndpoint
             return $this;
         }
 
-        $this->node_id = $node_id;
+        $this->node_id = urlencode($node_id);
 
         return $this;
     }
--- a/src/Elasticsearch/Endpoints/Cat/Fielddata.php
+++ b/src/Elasticsearch/Endpoints/Cat/Fielddata.php
@@ -28,7 +28,7 @@ class Fielddata extends AbstractEndpoint
             return $this;
         }
 
-        $this->fields = $fields;
+        $this->fields = urlencode($fields);
 
         return $this;
     }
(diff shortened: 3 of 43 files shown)


Nexucis:
Hi, my apologies to disturb you. I would like to know if this pull request will interest you :). Or maybe that's no up to date? 

Can you give me a feed back about this PR please? 
Thank you for your continuous work on this project!

polyfractal:
Hi, sorry for ignoring this... just fell off my radar.

I'm not sure the urlencoding is the issue to be honest.  There are tests that use unicode characters (e.g. https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/test/index/10_with_id.yml#L6) that work just fine.  That test would translate into a call to `POST /test-weird-index-中文/weird.type/1`

So there may be something else going on here.

That said, I'm not sure the client should support urlencoding anyway.  The low-level Java client doesn't, for example, and it was done that way on purpose:  https://github.com/elastic/elasticsearch/issues/24987#issuecomment-308097834

Lemme talk to the other client maintainers and get back to you.

Nexucis:
Hello, thanks for your feedback. 
I see the test that you'r linking, but I didn't see where there is a test with UTF-8 characters in alias. But maybe it is the same thing isn't it? 

In fact It been a while that I worked about this thing, I should test the issue with the last elasticsearch version. Maybe all problems with alias afe already fixed. 

However, the other issue that you link was very interesting. And now I think in the same way. But perhaps this url encoding needs to go in an helper. What do you think? 

And do you think it help you that I have extended all alias endpoint to the AbstractAliasEndpoint. I can do that in a different pull request if you want :)

Nexucis:
This night, I was thinking about something. In the php client, there is just the alias who is encoding not all the url like in the java client. 

So compared to what I said before, maybe we can just leave it like that because

polyfractal:
Aliases are normally passed as part of the URL, so it would be handled just like UTF8 in an index name.  What APIs were you having problems with in particular?  Can you recreate it with some test code that I can run?

Nexucis:
Ok I come back to you once I wrote the tests (if the problem is still there). 
Maybe I'll put the test in another PR in order to see the test's effect.

ezimuel:
Closing due to inactivity.

PR #484, opened 2016-10-26
Catch additional exceptions in the ping function for those who use mu…

josephtaylor-mn:
…ltiple nodes in their connection pool.

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -4,6 +4,7 @@ namespace Elasticsearch;
 
 use Elasticsearch\Common\Exceptions\BadMethodCallException;
 use Elasticsearch\Common\Exceptions\InvalidArgumentException;
+use Elasticsearch\Common\Exceptions\NoNodesAvailableException;
 use Elasticsearch\Common\Exceptions\Missing404Exception;
 use Elasticsearch\Common\Exceptions\TransportException;
 use Elasticsearch\Endpoints\AbstractEndpoint;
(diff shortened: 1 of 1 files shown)


josephtaylor-mn:
Does anyone have any insight into how to fix the issues being rose by the UnitTests? I am cannot find the point of failure and am wondering if it is perhaps not related to the code that I have altered.

polyfractal:
I think this is an unrelated failure, but I'm on mobile boarding a plane so
wasn't able to look closely.

I'll take a closer look tomorrow and get back to you :)

Thanks for the PR btw!

josephtaylor-mn:
Any news on the issues related to my PR not passing? Any help I can provide to get this merged in I would love to provide!!

Thanks!

polyfractal:
Argh, sorry, i was travelling when I first saw this, then went on vacation a few days later.  And completely forgot. :(

Is there a reason you added the BadRequest400 exception?  400-level errors are returned when the client sends a badly constructed request (e.g. bad query syntax), which generally indicates the server is working just fine.

polyfractal:
Re: those tests... I think that was transient test failures at the time (the runner was choking on certain responses due to some recent changes).  If you rebase onto the current master and re-push, we can check against the updated test code.

polyfractal:
I'm going to modify this slightly and remove the BadRequest400, and merge.  Thanks!

polyfractal:
Merged in: https://github.com/elastic/elasticsearch-php/commit/d857dd39523db62fd4307e5a3d902f6d35a8fc1e

PR #490, opened 2016-11-02
Update php-version-requirement.asciidoc

evertramos:
Does 5.0 requires only version PHP 5.4+


polyfractal:
Ah, actually, this page should be updated.  It's not wrong (ES-PHP 2.0 requires PHP 5.4+), but I forgot to update it for the new 5.0 branch.

Do you want to PR that, or should I add it?

polyfractal:
Aaaand just saw your other PR :)

evertramos:
Yeah sorry.. I am reading it all and just found that should be php 5.6

PR #492, opened 2016-11-02
Update crud.asciidoc

evertramos:
Add a 'pipeline' example in the docs when using ingest attachment.


polyfractal:
Gonna reject this, mainly because I think it'll confuse people.  Ingest pipelines are fairly niche at the moment, so including it in the general PHP docs will likely just confuse people.  Thanks though!

PR #502, opened 2016-11-17
Introducing constant JSON_FORCE_OBJECT to json_encode

### LINKED ISSUE #495: Issue with "match_all" search query
### Summary

With version 5.0.0, the `match_all` query is throwing `parsing_exception: [match_all] query malformed, no start_object after query name`. Executing a basic `match_all` query over REST API works as intended.

### Code snippet

```php
$params = [
    "index" => "foo",
    "type" => "bar",
    "query" => [
        "match_all" => [],
    ],
];
```

### Workaround
The issue seems to be that the `match_all` query has no parameters. If I set the `boost` parameter to `1.0` (the default behaviour), the query works.

```php
$params = [
    "index" => "foo",
    "type" => "bar",
    "query" => [
        "match_all" => [
            "boost" => 1.0,
        ],
    ],
];
```

### System details

- Operating System: Ubuntu 16.04.1 64-bit
- PHP Version: 7.0.8
- ES-PHP client version: 5.0.0
- Elasticsearch version: 5.0

kubawerlos:
- small fix to PHPUnit configuration
 - 2 tests to ensure nothing will be broken
 - serializers method serialize simplification and adding constant JSON_FORCE_OBJECT

It will resolve #495 and also make serializers to work on 2.* and 5.* versions of Elasticsearch.

### DIFF
--- a/phpunit-integration.xml
+++ b/phpunit-integration.xml
@@ -8,7 +8,7 @@
         verbose="true"
 >
     <php>
-        <env name="ES_TEST_HOST" value="http://localhost:9200"/>
+        <server name="ES_TEST_HOST" value="http://localhost:9200"/>
     </php>
     <testsuites>
         <testsuite>
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -8,7 +8,7 @@
         verbose="true"
 >
     <php>
-        <env name="ES_TEST_HOST" value="http://localhost:9200"/>
+        <server name="ES_TEST_HOST" value="http://localhost:9200"/>
     </php>
     <testsuites>
         <testsuite>
--- a/src/Elasticsearch/Serializers/ArrayToJSONSerializer.php
+++ b/src/Elasticsearch/Serializers/ArrayToJSONSerializer.php
@@ -22,16 +22,9 @@ class ArrayToJSONSerializer implements SerializerInterface
      */
     public function serialize($data)
     {
-        if (is_string($data) === true) {
-            return $data;
-        } else {
-            $data = json_encode($data, JSON_PRESERVE_ZERO_FRACTION);
-            if ($data === '[]') {
-                return '{}';
-            } else {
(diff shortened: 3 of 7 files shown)


polyfractal:
Admittedly, it's been a while since I've looked at this closely, but I believe this won't work for all of ES API either.  E.g. there are instances in the API where empty arrays are valid, but empty objects are not.  So forcing everything to an object will break that part of the API instead.

I believe I chose this route because the instances of empty objects are relatively few, and easier to mitigate than empty arrays.

kubawerlos:
Any example of such case? I would write a functional test to cover it end then maybe some simplification.

polyfractal:
Unsure, lemme dig through the API a bit.  It may no longer be an issue, ES 5.x did a lot of work cleaning up the API so that only one format is accepted (there used to be a lot more "magic" in the parsing to tolerate differences, which caused huge headaches).

It'd be great if we could enable this constant, I agree.  Will get back to you in a bit :)

ruflin:
@polyfractal @kubawerlos I just had a PR 2 days ago to switch to JSON_FORCE_OBJECTS and it broke quite a lot of things: https://github.com/ruflin/Elastica/pull/1219 For an indication where it doesn't work, check out the failing tests. It could also be that this is only related to Elastica ...

kubawerlos:
Thanks @ruflin, I guess this PR was way too optimistic.

polyfractal:
Took a slightly closer look and yeah, I think this will break things (unfortunately) :(

Off the top of my head, both function_score and indices will complain about empty objects for functions (`functions: {}`) and indices (`indices:{}`) respectively.  There are likely others, in particular I'm thinking highlighting and sorting, and perhaps some aggregations.

:(

Thanks for the PR anyhow!  Maybe we can revisit this in the future after some more cleanups on the ES side

Issue #382, opened 2016-01-25
Allow custom curl configuration for entire client


polyfractal:
Currently, you can configure custom curl options/headers on a per-request basis.  It would be convenient if the same could be specified during client creation, and applied to all requests.

Related #376

ride-user:
+1

rayward:
Would be as simple as allowing the connection params to be configured in the `ClientBuilder` rather than having it hardcoded as an empty array: https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/ClientBuilder.php#L428 ?

Oliboy50:
@rayward indeed
So, what is the main issue with that?

PR #506, opened 2016-11-29
add ClientBuilder->setConnectionParams()

Oliboy50:
We want to be able to pass `['client'][]` options globally instead of "per request".

This PR hopes to be merged in a 2.3.0 release (because I don't use 5.x yet). 👼 
See https://github.com/elastic/elasticsearch-php/pull/507 for `master` branch PR.

Related issues: https://github.com/elastic/elasticsearch-php/issues/382 https://github.com/elastic/elasticsearch-php/issues/259

Issue #259, opened 2015-06-30
How to configure curl connect timeout?


TheHett:
Hi,
I do not understand, how to pass "timeout" param to transport connection? Its possible?
Using param "timeout" in method "search" no give effect.

polyfractal:
Yeah, the search `timeout` param is the Elasticsearch timeout.  E.g. how long a node should wait for a shard response before returning.

If you want a curl connection timeout, you can do it two ways (assuming you are using the default Guzzle connection):

```php
$params['guzzleOptions']['command.request_options'] = [
    'connect_timeout' => 2.0,
    'timeout' => 2.0
];
$client = new Elasticsearch\Client($params);
```

Or raw curl options:

```php
$params['guzzleOptions'] = [
    'curl.options'=> [
        CURLOPT_CONNECTTIMEOUT => 2.0
    ]
];
$client = new Elasticsearch\Client($params);
```

Unfortunately, these can only be set when the connection is instantiated at the moment.  No technical reason, other than it hasn't been fixed to be per-request yet.  ES-PHP 2.0 allows per-request options (or at least it should...been a while since I've looked at it)

TheHett:
:+1: thank you very much

TheHett:
@polyfractal 
Hi, you are sure that here
https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Common/DICBuilder.php#L291
need use array_merge? I suspect it need array_merge_recursive

harpreetsb:
how to do this in ES2 syntax?

polyfractal:
@harpreetsb In ES-PHP 2.x the timeout is now specified per-request.  See the documentation here:  https://www.elastic.co/guide/en/elasticsearch/client/php-api/2.0/_per_request_configuration.html#_curl_timeouts

```php
$client = ClientBuilder::create()->build();

$params = [
    'index' => 'test',
    'type' => 'test',
    'id' => 1,
    'client' => [
        'timeout' => 10,        // ten second timeout
        'connect_timeout' => 10
    ]
];
$response = $client->get($params);
```

TheHett:
@polyfractal thx!

PR #507, opened 2016-11-29
add ClientBuilder->setConnectionParams()

Oliboy50:
Closes https://github.com/elastic/elasticsearch-php/issues/382 

We want to be able to pass `['client'][]` options globally instead of "per request".

This is the PR for the `master` branch, see https://github.com/elastic/elasticsearch-php/pull/506 for `2.x` branch.

Other related issues: https://github.com/elastic/elasticsearch-php/issues/259

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -71,6 +71,9 @@ class ClientBuilder
     /** @var array */
     private $hosts;
 
+    /** @var array */
+    private $connectionParams;
+
     /** @var  int */
     private $retries;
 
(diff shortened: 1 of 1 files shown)


johntiror:
Hi, I couldn't find documentation regarding this new method. How can I set the timeout? like this?
`$params['client']['timeout']=10;`

Thanks

polyfractal:
@johntiror Yep!

There's also `$params['client']['connect_timeout']` which controls the original connection timeout duration (`'timeout'` controls the request timeout, not the initial connection).  Most folks don't really care which, but sometimes the distinction matters :)

PR #511, opened 2016-12-11
Allow PHP 7.1 in composer enviroment

Haehnchen:


composer.json
```diff
@@ -10,7 +10,7 @@
     }
   ],
   "require": {
-    "php": "^5.6.6|^7.0",
+    "php": "^5.6.6|^7.*",
```
Oliboy50: I'd even write `^5.6.6||>=7.0`

polyfractal:
Yeah, I'm not sure this is needed?  [Composer docs say](https://getcomposer.org/doc/articles/versions.md#caret) the caret works for all "non-breaking" releases before the next significant version:

> The ^ operator behaves very similarly but it sticks closer to semantic versioning, and will always allow non-breaking updates. For example `^1.2.3` is equivalent to `>=1.2.3 <2.0.0` as none of the releases until 2.0 should break backwards compatibility.

So `^7.0` means all non-breaking releases before the next significant version (8.0), or `>=7.0, <8.0)

PR #514, opened 2016-12-12
Use `array_diff` to Check Endpoint Parameters

chrisguitarguy:
Slightly nicer than the old array_search method that iterates through every user provided param. Also gives the user feed back on *all* invalid parameters at once, rather than just one at a time.

Also adds a test for `AbstractEndpoint` to verify that the exception does actually happen.


polyfractal:
<3

PR #527, opened 2017-01-17
DeleteByQuery method access level fix

mxp100:


polyfractal:
Oops, that's my fault.  We re-committed this from an old release, which apparently was before the endpoint refactoring happened.

Thanks!  Merging :)

PR #538, opened 2017-02-08
JSON seriasers should throw exception on encode error

devanubis:
If request data is not properly encoded the serialisers can return boolean false

This eventually leads to an uncaugh InvalidArgumentException: "Invalid request body provided" from GuzzleHttp\Ring\Client\CurlFactory which is expecting a body.

IMHO throwing an exception early would be preferable, as it should explain why the body is not correct and make things easier for developers to find their code which is not encoding the body correctly.

### DIFF
--- a/src/Elasticsearch/Serializers/ArrayToJSONSerializer.php
+++ b/src/Elasticsearch/Serializers/ArrayToJSONSerializer.php
@@ -2,6 +2,8 @@
 
 namespace Elasticsearch\Serializers;
 
+use Elasticsearch\Common\Exceptions\RuntimeException;
+
 /**
  * Class JSONSerializer
  *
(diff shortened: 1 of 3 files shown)


polyfractal:
++  Could you add it to the SmartSerializer too?

Also, would you mind [Signing the CLA](https://www.elastic.co/contributor-agreement/) so that we can merge this?

karmi:
Hi @devanubis, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/elasticsearch-php/pull/538.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

devanubis:
Hi, I've rebased my commit to include SmartSerializer too (sorry didn't realise that also did JSON encoding) and I've also amended my author details (old config file on another box)

karmi:
@devanubis, green for the CLA check now, thanks!

polyfractal:
Thanks @devanubis!  Merging :)

Jalle19:
Any chance this could be backported to the 5.x branch? I recently stumbled upon this issue and it took me 30 minutes to realize it was the serializer that failed to encode the request body.

PR #542, opened 2017-02-17
Make id param optionnal for termvectors requests.

afoucret:
ES REST API allow usage of the termvectors API to process document that are not existing into an index as explained in the Example 3. Artificial documents of the termvectors documentation page (https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-termvectors.html).

The PHP client is not consistent with the REST API since the id parameter is mandatory.
This PR allows to check the presence of an id or of a document and raise an exception if neither of those parameters is present.

It was not required until recently (<5.x) since we were to hack this able to specifiy a empty string as an id.

### DIFF
--- a/src/Elasticsearch/Endpoints/TermVectors.php
+++ b/src/Elasticsearch/Endpoints/TermVectors.php
@@ -39,27 +39,31 @@ class TermVectors extends AbstractEndpoint
     public function getURI()
     {
         if (isset($this->index) !== true) {
-            throw new Exceptions\RuntimeException(
-                'index is required for TermVectors'
-            );
-        }
-        if (isset($this->type) !== true) {
-            throw new Exceptions\RuntimeException(
-                'type is required for TermVectors'
-            );
-        }
-        if (isset($this->id) !== true) {
-            throw new Exceptions\RuntimeException(
-                'id is required for TermVectors'
-            );
-        }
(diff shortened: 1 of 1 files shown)


polyfractal:
Right you are, thanks!

We normally shy away from any body validation, but I think it makes sense in this situation.  Merging!

afoucret:
Hi @polyfractal,

Do you have a timeframe for the next release ? Will this one being part of the release ?

But huge fav of using master in my dependencies.

polyfractal:
Nothing planned, but I'd be fine cutting a bug-release for this.  I need to wrap up a few things first, but I'll cut it by the end of this afternoon. :)

afoucret:
Thank you for being so quick to respond. I stay tuned. 

BR,

afoucret:
Hi @polyfractal,

I did not see any new release containing this PR. Is there any specific issue on having a bugfix release now ?

BR,

afoucret:
Hi @polyfractal,

Any news about this one ?

polyfractal:
I'm so sorry this took forever to be released.  I've been heads-down helping the ML team integrate with ES in 5.4, which has kept me quite busy.

I just tagged ES-PHP `v5.2.0`, which has this PR.  Sorry again! :(

PR #545, opened 2017-02-21
bool query's "query" should use a "should"

sarwarbhuiyan:
replaced query with "should" in bool query


polyfractal:
Whoops :)  Thanks

PR #549, opened 2017-02-25
Fall back for the empty array as new std class

vrkansagara:
fall back for the empty array as new std class

### DIFF
--- a/docs/search-operations.asciidoc
+++ b/docs/search-operations.asciidoc
@@ -236,7 +236,7 @@ $params = [
     "index" => "my_index",
     "body" => [
         "query" => [
-            "match_all" => []
+            "match_all" => [] // or use new \stdClass() 
         ]
     ]
 ];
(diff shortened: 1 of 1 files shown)


polyfractal:
Actually, how about we just replace that `[]` with the stdClass so it is unambiguous which should be used?

vrkansagara:
`stdClass` work in all the case with any library version and matter of the nosql I like object thinking not `[]`  thinking. and class and object I like most because It give flexibility to do more operation on it.

polyfractal:
@vrkansagara Right, I meant instead of:

```php
 "match_all" => [] // or use new \stdClass() 
```
we just do:
```php
 "match_all" => new \stdClass() 
```

:)

vrkansagara:
yes @polyfractal  that Want I like. using class notation is very much clear to understand.

polyfractal:
Want to update your PR so it uses that notation, and I'll merge it?

vrkansagara:
@polyfractal  done with PR updation. 👍

polyfractal:
Thanks!  Merging :)

PR #551, opened 2017-02-28
added getTransport, getEndpoint, getRegisteredNamespacesBuilder to Cl…

MichaelBrenden:
…ientBuilder.php

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -98,6 +98,36 @@ class ClientBuilder
     }
 
     /**
+     * Can supply first parm to Client::__construct() when invoking manually or with dependency injection
+     * @return this->ransport
+     *
+     */
+    public function getTransport()
+    {
+        return $this->transport;
+    }
+
+    /**
+     * Can supply second parm to Client::__construct() when invoking manually or with dependency injection
+     * @return this->endpoint
+     *
+     */
+    public function getEndpoint()
+    {
+        return $this->endpoint;
+    }
+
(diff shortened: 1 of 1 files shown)


polyfractal:
Well, I don't have any particular opposition to merging this since they are harmless, although I'm not sure I see exactly what you're trying to accomplish.  E.g. the `ClientBuilder` generates the `Client`, there shouldn't be any need to directly instantiate a `Client` yourself (and so you shouldn't need to get a `Transport`, etc).

Also, some things like `Transport` are usually null until `build()` is invoked, at which point they are constructed because they depend on other objects which are only guaranteed to exist during the build process.

That said, if this makes your life easier, I don't see why the changes shouldn't be merged. :)  Would you mind [Signing the CLA](https://www.elastic.co/contributor-agreement/)?

MichaelBrenden:
CLA signed.

MichaelBrenden:
CLA signed.

MichaelBrenden:
ClientBuilder generates the Client, yes, but Dice needs (or at least prefers) to be able to generate the Client.  So we use ClientBuilder, run its build(), then pull the parts out that Client needs, and let Dice generate the Client, feeding it the needed pieces from ClientBuilder.  Another way (#2) that Dice can work is to have dependency-inject the ClientBuilder, but then each child-class' __construct would have to do a $this->Client = $this->ClientBuilder->build().  So doing it the first way is easier.  Thank you for looking.

polyfractal:
Fair enough.  Thanks for signing the CLA, merging! :)

PR #557, opened 2017-03-20
endpoint index and type getters

ewgRa:
1. Is it possible to add getIndex and getType getters for AbstractEndpoint?
If not, what is a purpose to make it so private?

We need it for https://github.com/ruflin/Elastica/pull/1275

2. Also, why:
```
        public function setType($type)
    {
        if ($type === null) {
            return $this;
        }


```
so strange? If I want drop type for my endpoint, why client force me to recreate full endpoint? Or at least why it is so silent?

### DIFF
--- a/src/Elasticsearch/Endpoints/AbstractEndpoint.php
+++ b/src/Elasticsearch/Endpoints/AbstractEndpoint.php
@@ -96,6 +96,14 @@ abstract class AbstractEndpoint
     }
 
     /**
+     * @return string|null
+     */
+    public function getIndex()
+    {
+        return $this->index;
+    }
+
+    /**
      * @param string $index
      *
      * @return $this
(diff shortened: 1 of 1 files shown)


polyfractal:
Hi!  Sorry for the delay, I've been sick this week :(

> Is it possible to add getIndex and getType getters for AbstractEndpoint? If not, what is a purpose to make it so private?

++!  There's no particular reason they didn't exist.  The client didn't need the functionality previously, but I don't see any problem adding the accessors.

> so strange? If I want drop type for my endpoint, why client force me to recreate full endpoint? Or at least why it is so silent?

Historical, mostly.  In the beginning, it was an attempt to help safety-proof the client against bad input. Over time my views on that have changed, but the code remained the same.  And no one has ever requested the ability to "unset" a param/type on an endpoint (remarkably enough), so this code has remained.

We can certainly remove these checks, I don't have a strong opinion about keeping them.

Sorry again fro the delay!

polyfractal:
PR LGTM, merging!

ewgRa:
Thanks!

PR #565, opened 2017-04-04
Remove incorrect Composer instructions from README

curry684:
Instructions on using the `--no-dev` flag were wrong as Composer will in this case never install the dev dependencies anyway. Dev dependencies are only ever resolved in the root package of an application for exactly the reasons that were described here. A consumer of the library will therefore never get the dev requirements of any library, with any flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ The recommended method to install _Elasticsearch-PHP_ is through [Composer](http
 3. Install your dependencies:
 
     ```bash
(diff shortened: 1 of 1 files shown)


curry684:
CLA was signed 3 seconds before submitting the PR, apparently there is a small delay  ;)

polyfractal:
TIL :) 

Thanks for the PR!

Issue #564, opened 2017-04-03
Scroll returning "Unrecognized token"


marksparrish:
This is on elasticsearch 5.3

When executing the scroll (after the search) I'm getting an "Unrecognized token" error.

Looking through the code, I think endpoint now wants the scroll_id key => value in the body.

https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html#sliced-scroll

POST /_search/scroll
{
    "scroll" : "1m", 
    "scroll_id" : "$$28$$$$i9puqxqirM3_lZ90VILTFnLaSJ4=DnF1ZXJ5VGhlbkZldGNoBQAAAAAAAOS1FmJsbWlkOXZOUnM2d3hwS05GUTJMWmcAAAAAAADktxZibG1pZDl2TlJzNnd4cEtORlEyTFpnAAAAAAAA5LgWYmxtaWQ5dk5SczZ3eHBLTkZRMkxaZwAAAAAAAOS2FmJsbWlkOXZOUnM2d3hwS05GUTJMWmcAAAAAAADkuRZibG1pZDl2TlJzNnd4cEtORlEyTFpn"
}

It looks like when it extracts the scroll_id is does not add the "scroll_id" key and just adds the scroll_id value.

The function in Elasticsearch/Endpoints/Scroll.php should be updated to:

    /**
     * @param $scroll_id
     *
     * @return $this
     */
    public function setScrollId($scroll_id)
    {
        if (isset($scroll_id) !== true) {
            return $this;
        }

        $this->body = ['scroll_id' => $scroll_id];

        return $this;
    }

polyfractal:
Yep, you're exactly right, that's the problem.  I fixed this on my local branch over the weekend... but I have one more test failure I want to fix, then I'll get it pushed and released.  Will do my best to release it tomorrow morning, sorry!

tomsommer:
Any chance of a release, please?

erichard:
I found a workaround. You need to pass the `scroll_id` in the `body` like this:

```
$client->scroll([
    'scroll_id' => $scrollId,
    'scroll' => '5m',
    'body' => [
        'scroll_id' => $scrollId,
    ]
]);
```

polyfractal:
[...]

PR #568, opened 2017-04-06
Fixed scroll example

pborreli:
Fixing #564 for branch 5.0+


polyfractal:
Thanks!  This was actually just broken, and should be fixed in https://github.com/elastic/elasticsearch-php/commit/ef0a3d204cff5a339159431b1aa42fb16d56450d.  I'm going to close this PR because the docs were correct... just the code was broken :)

PR #569, opened 2017-04-11
Client::exists() should not throw a warning when a resource does not exist

rgeraads:
I made a change in the `exist()` method, it should not trigger a warning when the resource could not be found. Because the sole purpose of `exist()` is that it checks for this.

Therefore that method should ignore 404 response errors.

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -559,6 +559,9 @@ class Client
         //manually make this verbose so we can check status code
         $params['client']['verbose'] = true;
 
+        //ignore warning when resource does not exist
+        $params['client']['ignore'] = ['404'];
+
         /** @var callback $endpointBuilder */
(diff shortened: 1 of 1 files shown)


polyfractal:
I don't think this is needed?  The Exists call is wrapped in a BooleanRequestWrapper, which [automatically converts all non-200 requests into a `false`](https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Namespaces/BooleanRequestWrapper.php#L43-L47).  

Were you seeing a specific exception that wasn't being caught?  

Note: if you're using the async mode, we can't do this wrapping easily so it's up to the user.

rgeraads:
It returns false indeed in case the document does not exist, but it also generates a warning message because the document does not exist.

ezimuel:
Closing due to inactivity.

PR #570, opened 2017-04-13
Fixed missing "body" in example.

DracoBlue:
The example to scroll for php is wrong atm. This PR fixes the example.


polyfractal:
Did you try the latest tagged release, `v5.2.0`?  That has fixes for scrolling... adding the body param shouldn't be necessary, it was just a bug :)

DracoBlue:
Nope. I can try later! But if it is not necessary to use it, we could remove it from the docs, too ;).

ezimuel:
Closing due to inactivity.

PR #576, opened 2017-05-02
Typos

loekvangool:

### DIFF
--- a/docs/index-operations.asciidoc
+++ b/docs/index-operations.asciidoc
@@ -95,7 +95,7 @@ $params = [
                     'reuters' => [
                         'type' => 'custom',
                         'tokenizer' => 'standard',
-                        'filter' => [ 'lowercase', 'stop', 'kstem']
+                        'filter' => ['lowercase', 'stop', 'kstem']
                     ]
                 ]
             ]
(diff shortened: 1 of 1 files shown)


docs/index-operations.asciidoc
```diff
@@ -240,16 +240,16 @@ The Get Mappings API will return the mapping details about your indexes and type
 $response = $client->indices()->getMapping();
 
 // Get mappings for all types in 'my_index'
-$params = ['index' => 'my_index'];
+$params = [ 'index' => 'my_index'];
 $response = $client->indices()->getMapping($params);
 
 // Get mappings for all types of 'my_type', regardless of index
-$params = ['type' => 'my_type' ];
+$params = [ 'type' => 'my_type' ];
```
Oliboy50: you could also remove the last space after `my_type` then

loekvangool: That would not make the file consistent yet. Do you know the convention? I will update the file to match it.

polyfractal: I don't think there is a convention, but I'd prefer it to at least be consistent within the page.  E.g. either spaces before/after, or no spaces at all.

I think I'd lean towards no spaces before/after personally.

ezimuel:
Closing due to inactivity.

PR #607, opened 2017-07-26
Change inheritance of 'NoNodesAvailableException' to 'ServerErrorResp…

xserrat:
Hi,

I've seen that `NoNodesAvailableException` extends from generic `\Exception` instead of `ServerErrorResponseException`. That's why I want to request information about client only if the engine will respond using `ElasticSearch\Client::info()` and I can't do it if an error on server exists.

Thank you!


ezimuel:
@xserrat can you sing the CLA and update the PR with the `master`? Thanks.

xserrat:
Done @ezimuel !

ezimuel:
Thanks @xserrat and sorry for the delay of my reply.

PR #609, opened 2017-08-01
[TEST] add newer ES versions to Travis build matrix

mhujer:
This PR adds newer Elasticsearch versions to the TravisCI build matrix.

The tests are failing for ES 5.6 (unreleased?) - which is used as a default for the builds without the ES version specified. It probably does not block the merge of the PR as the tests would start failing for any other PR as well.

Maybe the Travis configuration can be changed so it is always running the builds against specific version (so the tests won't start failing after the new version is released). <-- What do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,28 +14,31 @@ branches:
 
 sudo: true
 
-php:
-  - 5.6
-  - 7.0
-  - 7.1
-  - hhvm
-
 matrix:
   fast_finish: true
   include:
+    - php: 5.6
+      env: ES_VERSION="5.0" TEST_BUILD_REF="origin/5.0"
+    - php: 5.6
+      env: ES_VERSION="5.5" TEST_BUILD_REF="origin/5.5"
+
+    - php: 7.0
+      env: ES_VERSION="5.0" TEST_BUILD_REF="origin/5.0"
+    - php: 7.0
+      env: ES_VERSION="5.5" TEST_BUILD_REF="origin/5.5"
+
     - php: 7.1
       env: ES_VERSION="5.0" TEST_BUILD_REF="origin/5.0"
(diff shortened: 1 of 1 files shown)


mhujer:
I've added second commit which specifies all possible combinations to build and excludes the ones that are failing

polyfractal:
Thanks for this! I had indeed forgotten to update the test matrix :) Would you mind PRing this against the `5.0` branch though?  Master cut over to Elasticsearch 6.x a week or two ago (not documented anywhere, sorry!), so we'll actually need to update travis to start targeting 6.x on master.

Regarding travis in general, well, it's been a rocky relationship :)  It's not always been easy to keep travis happy with the test framework -- likely my own inexperience with Travis --  so keeping it up to date has been lower on my priority list.

I've been generally trying to keep the test matrix to "just a few" versions of Elasticsearch.  Mostly to be nice to Travis... the tests take quite some time to run, and there are not generally breakages with old versions that don't also show up in new versions.  So I don't think we need the full 5.x series, just the latest few

> Maybe the Travis configuration can be changed so it is always running the builds against specific version (so the tests won't start failing after the new version is released). <-- What do you think?

I'm not entirely sure I know what you mean here?

mhujer:
I've create a new PR with ES 6 build: #612

mhujer:
I've changed the target branch of this PR to `5.0`.

The build matrix currently builds all combinations of:
- PHP 5.6, 7.0 and 7.1
- ES 5.0 and 5.5 (to check that the library is compatible with the oldest and newest ES version in 5.x branch)

(6 builds in total + 1 for HHVM):

polyfractal:
LGTM!  Thanks and sorry for the delay on this, has been a busy week :)

Issue #611, opened 2017-08-05
Consider dropping HHVM support


mhujer:
I propose removing HHVM from the TravisCI build matrix because:
- The 5.0 build does [not run on HHVM at all](https://travis-ci.org/elastic/elasticsearch-php/jobs/255754000).
- The library itself probably [cannot be installed in HHVM (Travis' Ubuntu Trusty)](https://travis-ci.org/mhujer/elasticsearch-php/jobs/261278633)
- not many people [use HHVM anyway](https://seld.be/notes/php-versions-stats-2017-1-edition) (from those who use Composer)

Big frameworks are also dropping support for HHVM: 
- Symfony (https://symfony.com/blog/symfony-4-end-of-hhvm-support)
- Zend Framework (https://framework.zend.com/blog/2017-06-06-zf-php-7-1.html)
- Doctrine (https://github.com/doctrine/doctrine2/issues/6424)

What is your opinion on this @polyfractal ?

polyfractal:
:+1: I'm for it.  I haven't talked to anyone using ES-PHP with HHVM, either in person or via an issue.  And it's never been a priority to support (obviously, given the failing build :) )

Let's drop it.

mhujer:
I've created PRs for `5.0` (#615) and for `master` (#616)

polyfractal:
Merged!

PR #612, opened 2017-08-05
[TEST] build against ES 6 on Travis

mhujer:
As mentioned in #609, the master should be tested against ES 6.x

- I've kept the 5.5 version there. So it can be easily seen, when the 5.x compatibility is broken. I think it should be kept there as it would allow the users to upgrade the elastisearch-php first and their ES cluster later.

TODO:
- [ ] Decide if the new version of elastisearch-php should support ES 5.x
- [x] Decide the fate of HHVM support: https://github.com/elastic/elasticsearch-php/issues/611
- [x] Merge https://github.com/elastic/elasticsearch-php/pull/616

EDIT:
The build fails for 6.0 which I think is OK as there are probably some changes required in `elastisearch-php` to fully support it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ sudo: true
 php:
   - 7.0
   - 7.1
-  - hhvm
 
 matrix:
   fast_finish: true
@@ -28,8 +27,6 @@ matrix:
       env: ES_VERSION="5.1" TEST_BUILD_REF="origin/5.1"
     - php: 7.1
       env: ES_VERSION="5.2" TEST_BUILD_REF="origin/5.2"
-  allow_failures:
-    - php: hhvm
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,24 +14,28 @@ branches:
 
 sudo: true
 
-php:
-  - 7.0
-  - 7.1
-
 matrix:
   fast_finish: true
   include:
+    - php: 7.0
+      env: ES_VERSION="5.5" TEST_BUILD_REF="origin/5.5"
+    - php: 7.0
+      env: ES_VERSION="6.0" TEST_BUILD_REF="origin/6.0"
+    - php: 7.0
+      env: ES_VERSION="6.x" TEST_BUILD_REF="origin/6.x"
+
     - php: 7.1
-      env: ES_VERSION="5.0" TEST_BUILD_REF="origin/5.0"
+      env: ES_VERSION="5.5" TEST_BUILD_REF="origin/5.5"
     - php: 7.1
-      env: ES_VERSION="5.1" TEST_BUILD_REF="origin/5.1"
+      env: ES_VERSION="6.0" TEST_BUILD_REF="origin/6.0"
     - php: 7.1
-      env: ES_VERSION="5.2" TEST_BUILD_REF="origin/5.2"
+      env: ES_VERSION="6.x" TEST_BUILD_REF="origin/6.x"
+
+  allow_failures:
+    - env: ES_VERSION="6.x" TEST_BUILD_REF="origin/6.x"
 
 env:
   global:
-    - ES_VERSION="5.x"
-    - TEST_BUILD_REF="origin/5.x"
     - ES_TEST_HOST=http://localhost:9200
     - JAVA_HOME="/usr/lib/jvm/java-8-oracle/jre"
 
--- a/tests/Elasticsearch/Tests/YamlRunnerTest.php
+++ b/tests/Elasticsearch/Tests/YamlRunnerTest.php
@@ -56,6 +56,9 @@ class YamlRunnerTest extends \PHPUnit_Framework_TestCase
 
     /** @var array A list of skipped test with their reasons */
     private static $skippedTest = [
+        'cat.aliases/20_headers.yml' => 'Using java regex fails in PHP',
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -14,27 +14,35 @@ branches:
...
-      env: ES_VERSION="5.0" TEST_BUILD_REF="origin/5.0"
+      env: ES_VERSION="5.5" TEST_BUILD_REF="origin/5.5"
     - php: 7.1
-      env: ES_VERSION="5.1" TEST_BUILD_REF="origin/5.1"
+      env: ES_VERSION="6.0" TEST_BUILD_REF="origin/6.0"
     - php: 7.1
-      env: ES_VERSION="5.2" TEST_BUILD_REF="origin/5.2"
+      env: ES_VERSION="6.x" TEST_BUILD_REF="origin/6.x"
+
+    - php: hhvm
```
mhujer: https://github.com/elastic/elasticsearch-php/issues/611 needs to be resolved first

.travis.yml
```diff
@@ -14,27 +14,36 @@ branches:
...
+      env: ES_VERSION="5.5" TEST_BUILD_REF="origin/5.5"
+    - php: hhvm
+      env: ES_VERSION="6.0" TEST_BUILD_REF="origin/6.0"
+    - php: hhvm
+      env: ES_VERSION="6.x" TEST_BUILD_REF="origin/6.x"
+
   allow_failures:
     - php: hhvm
+    - env: ES_VERSION="6.x" TEST_BUILD_REF="origin/6.x"
```
mhujer: This is related to my previous comment:
> Maybe the Travis configuration can be changed so it is always running the builds against specific version (so the tests won't start failing after the new version is released). <-- What do you think?

Running against `6.x` means that the build can start failing without any changes in `elasticsearch-php` (because something was changed in the Elasticsearch itself).

I suggest the following:
- let the build run against `ES_VERSION="6.x"`
- put it into `allow_failures` so it won't break the build unexpectedly 
- when the new ES version is branched, it should be added to the build matrix (and checked that the build passes)

mhujer:
I realized that when the build is ran against ES 5.x, it won't the REST-tests through `YamlRunner` because if the rename from `yaml` to `yml` (in https://github.com/elastic/elasticsearch-php/commit/ceac5bd132a85bc51c19d4083e84becf6a688627).

It is fixed in second commit. (can be removed if you decide to build the master only against 6.x)

.travis.yml
polyfractal: Ah, I see what you mean.  I'm of split opinion: having builds fail on 6.x because of a new feature, change in API, etc is a convenient way to know that upstream has added something.  That said, I agree it can be frustrating for outsiders because PRs will fail for unrelated reasons.

Let's go ahead and mark 6.x as `allow_failures` :)

polyfractal:
Left a comment, and replied to the HHVM issue.  

Re: 6.0 build failure, I think there are still some lingering fixes in core that we're waiting on. I'll have another look at the tests soon.  We just released Elasticsearch 6.0.0-beta1, so it's a priority to get our 6.0 branch cleaned up and tagged for release too :)

.travis.yml
mhujer: I've removed the hhvm from the build matrix

mhujer:
1) HHVM removal resolved in [separate PR](https://github.com/elastic/elasticsearch-php/pull/616) (it should be merged first).

2) What is your opinion on this?
> I've kept the 5.5 version there. So it can be easily seen, when the 5.x compatibility is broken. I think it should be kept there as it would allow the users to upgrade the elastisearch-php first and their ES cluster later.

3) I've added a workaround for running both `.yml` and `.yaml` test files in `YamlRunnerTest` (to support 5.5)
4) I had to skip `cat.aliases/20_headers.yml` to make the 5.5 build pass. It looked like another regex incompatibility - see the build output https://travis-ci.org/mhujer/elasticsearch-php/jobs/263631465#L845 (Do you think it is OK to skip it for now?)

polyfractal:
> I've kept the 5.5 version there. So it can be easily seen, when the 5.x compatibility is broken. I think it should be kept there as it would allow the users to upgrade the elastisearch-php first and their ES cluster later.

I'm ok with minor fixes to keep 6.x compatible with 5.x, such as the `.yml`/`.yaml` change.  But we don't need to bend over backwards to keep them compatible if there is a big or otherwise obnoxious change required.  Since Elastic stack has moved to unified releases, all components are expected to be upgraded at the same time. 

The clients aren't officially part of that unified release process (we version independently), but the sentiment carries over for the most part.

So if we can keep compatibility, great.  Otherwise it's not a huge problem.  It's not generally an issue though, since the REST api doesn't dramatically change very often so even the 6.x client is mostly compatible with 1.x :)

> I had to skip cat.aliases/20_headers.yml to make the 5.5 build pass

That's fine, the cat tests in particular are super troublesome.  They are mainly for human consumption anyhow, so I've always deprioritized them over other issues :)

polyfractal:
Merged, after fixing up some conflicts.  I removed the `allow_failure` on 6.x, since we shouldn't be failing on that (and thus shouldn't tolerate failures on travis). :)

mhujer:
Perfect, thanks for clarifying the backwards compatibility policy!

PR #613, opened 2017-08-05
[Docs] recommend composer/ca-bundle instead of Kdyby/CurlCaBundle

mhujer:
Kdyby/CurlCaBundle was deprecated after [composer/ca-bundle](https://github.com/composer/ca-bundle) was released.


polyfractal:
:+1:

PR #617, opened 2017-08-11
Fix RoundRobinSelector

mhujer:
When working on some tests cleanup, I realized that there is an issue with `RoundRobinSelectorTest`. It was wasn't actually testing anything, because there was `assertEquals()` used to compare different instances of same class. It compares with `==` which means that different instances with same data are seen as same.

See https://3v4l.org/6Qipd

You can try it yourself, simply by changing `assertEquals()` to `assertSame()` in those two tests.

I changed several things - see the comments directly in the "Diff" tab.

### DIFF
--- a/src/Elasticsearch/ConnectionPool/Selectors/RoundRobinSelector.php
+++ b/src/Elasticsearch/ConnectionPool/Selectors/RoundRobinSelector.php
@@ -29,8 +29,10 @@ class RoundRobinSelector implements SelectorInterface
      */
     public function select($connections)
     {
+        $returnConnection = $connections[$this->current % count($connections)];
+
         $this->current += 1;
 
-        return $connections[$this->current % count($connections)];
+        return $returnConnection;
     }
 }
--- a/tests/Elasticsearch/Tests/ConnectionPool/Selectors/RoundRobinSelectorTest.php
+++ b/tests/Elasticsearch/Tests/ConnectionPool/Selectors/RoundRobinSelectorTest.php
@@ -5,6 +5,7 @@ declare(strict_types = 1);
 namespace Elasticsearch\Tests\ConnectionPool\Selectors;
 
 use Elasticsearch;
+use Elasticsearch\Connections\ConnectionInterface;
 
 /**
  * Class SnifferTest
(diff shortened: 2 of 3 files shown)


tests/Elasticsearch/Tests/ConnectionPool/Selectors/RoundRobinSelectorTest.php
```diff
@@ -29,19 +30,32 @@ public function testTenConnections()
     {
         $roundRobin = new Elasticsearch\ConnectionPool\Selectors\RoundRobinSelector();
 
-        $mockConnections = array();
-        foreach (range(0, 10) as $index) {
```
mhujer: `range(0, 10)` actually creates array with 11 items - see https://3v4l.org/kUBhi

tests/Elasticsearch/Tests/ConnectionPool/Selectors/RoundRobinSelectorTest.php
```diff
@@ -29,19 +30,32 @@ public function testTenConnections()
...
-            $mockConnections[$index] = $this->getMockBuilder('\Elasticsearch\Connections\CurlMultiConnection')
+        $mockConnections = [];
+        foreach (range(0, 9) as $index) {
+            $mockConnections[$index] = $this->getMockBuilder(ConnectionInterface::class)
```
mhujer: I replaced the string classname (btw. non-existent since ES-PHP 2.0 ;-) ) with `::class` reference. It will fail as soon as you remove the class.

tests/Elasticsearch/Tests/ConnectionPool/Selectors/RoundRobinSelectorTest.php
```diff
@@ -29,19 +30,32 @@ public function testTenConnections()
...
+        $this->assertSame($mockConnections[9], $roundRobin->select($mockConnections));
+
+        // select five - should start from the first one (index: 0)
+        $this->assertSame($mockConnections[0], $roundRobin->select($mockConnections));
+        $this->assertSame($mockConnections[1], $roundRobin->select($mockConnections));
+        $this->assertSame($mockConnections[2], $roundRobin->select($mockConnections));
+        $this->assertSame($mockConnections[3], $roundRobin->select($mockConnections));
+        $this->assertSame($mockConnections[4], $roundRobin->select($mockConnections));
 
-            $nextIndex = ($index % 10) + 1;
```
mhujer: I replaced the `foreach` with explicit calls, because I believe that the original test logic was broken:
1) The first `$index` is `0`, so the mod returns `0`
2) add 1 to it
3) select `$mockConnections[1]` <- not the first connection

src/Elasticsearch/ConnectionPool/Selectors/RoundRobinSelector.php
```diff
@@ -29,8 +29,10 @@ class RoundRobinSelector implements SelectorInterface
      */
     public function select($connections)
     {
+        $returnConnection = $connections[$this->current % count($connections)];
```
mhujer: This change means, that it would select the first connection (with the array index `0`). Before it selected the second connection (with the array index `1`)

tests/Elasticsearch/Tests/ConnectionPool/Selectors/RoundRobinSelectorTest.php
```diff
@@ -52,33 +66,39 @@ public function testTenConnections()
...
                 ->getMock();
         }
 
-        foreach (range(0, 4) as $index) {
-            $retConnection = $roundRobin->select($mockConnections);
-
-            $nextIndex = ($index % (count($mockConnections)-1)) + 1;
-            $this->assertEquals($mockConnections[$nextIndex], $retConnection);
-        }
+        // select five
```
mhujer: Same as above - I replaced quite complicated (and possibly broken) test logic with a straightforward code.

src/Elasticsearch/ConnectionPool/Selectors/RoundRobinSelector.php
polyfractal: Doesn't particularly matter since the connections are shuffled at the start, but can't hurt to fix while we're visiting the code :)

tests/Elasticsearch/Tests/ConnectionPool/Selectors/RoundRobinSelectorTest.php
polyfractal: ++ explicit is fine in my book. Easier to read

polyfractal:
LGTM, thanks!  I had no idea `assertEquals` used `==` instead of `===`.  Seems like that's quite trappy behavior... :(

Merging!

PR #618, opened 2017-08-11
[TEST] tests code cleanup

mhujer:
~~Depends on https://github.com/elastic/elasticsearch-php/pull/612 (should be merged first)~~

- [x] ~~merge 612~~

----

Reasoning for each change is included in the individual commit messages.

### DIFF
--- a/tests/Elasticsearch/Tests/ClientTest.php
+++ b/tests/Elasticsearch/Tests/ClientTest.php
@@ -6,6 +6,7 @@ namespace Elasticsearch\Tests;
 
 use Elasticsearch;
 use Elasticsearch\ClientBuilder;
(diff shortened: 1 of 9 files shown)


mhujer:
@polyfractal I've rebased this and it is ready for review.

polyfractal:
LGTM, thanks for these :)

Looking through the affected code, it reminds me that there are still a few unit tests that still rely on an externally running cluster (e.g. `SniffingConnectionPoolIntegrationTest()` does a ping, `testCustomQueryParams()` does an exists call) which is obviously not great for unit tests.  Should be mocked out instead.

I'll open an issue so that these don't continue being forgotten about :)

PR #619, opened 2017-08-12
[DOCS] Remove Sami from composer.json and update docs

mhujer:
Sami was somehow broken and I realized that they state in [the docs](https://github.com/FriendsOfPHP/Sami#installation), that it should not be installed as dependency in `composer.json`.

So I removed it from there and updated the script to download the .phar before rebuilding the docs.


polyfractal:
:+1:

PR #622, opened 2017-08-20
Travis: add PHP 7.2 + ES 6.0 to build matrix

mhujer:
PHP 7.2 is already quite stable (they have released beta3, I'm using it locally for day-to-day work with no issues) so I suggest adding it to Travis matrix. 

(The build against ES 6.0 fails, but the failing tests are the same as on 7.1)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,9 @@ matrix:
     - php: 7.1
       env: ES_VERSION="6.x" TEST_BUILD_REF="origin/6.x"
 
+    - php: 7.2
+      env: ES_VERSION="6.0" TEST_BUILD_REF="origin/6.0"
+
 env:
   global:
     - ES_TEST_HOST=http://localhost:9200


polyfractal:
++ I'm for it.  If for some reason it becomes noisy before release, we can `allow_failures`.

As an aside, I'm going to cut a 6.0 branch and tag a beta release shortly, just so you know :)

mhujer:
@polyfractal After you branch 6.0, I will target the `master` with my changes, right? And if some of them are worth backporting to 6.0, it can be done in separate PR.

If I understand it correctly, the `6.0` branch would be more like `6.x`, as it will be used for "fixes" and eventually releases `6.0`, `6.1`, `6.2` etc.?

polyfractal:
Just branched 6.0, and tagged `v6.0.0-beta1`.  Thanks for all the work on cleaning up the tests!

>@polyfractal After you branch 6.0, I will target the master with my changes, right? And if some of them are worth backporting to 6.0, it can be done in separate PR.
>
>If I understand it correctly, the 6.0 branch would be more like 6.x, as it will be used for "fixes" and eventually releases 6.0, 6.1, 6.2 etc.?

Yep, you're correct.  Target `master`, backport to `6.0`.  You can send a second PR, or I can just cherry-pick them over, whichever you prefer.

You're also correct, I treat the `6.0` branches as `6.x`, which obtains fixes and enhancements to go into the next minor.  I haven't bothered maintaining individual branches (`6.0`, `6.1`, etc) like Elasticsearch does because we rarely have bugfixes that need to be applied to a specific branch but not others... so it's easier to have just a single main branch for the major version and tell people to upgrade if they need to get around a bug.

At some point we'll want to cut `master` over to tracking Elasticsearch `master` (aka `7.0`), but that isn't necessary yet and would entail some changes to the travis config too.

PR #627, opened 2017-08-22
[TEST] PHPUnit 6.3.0

mhujer:
I've updated PHPUnit to 6.3.0.

I've done it in two steps (upgrading to 5.7.21, fixing configs and tests and then upgrading to 6.3.0), but it may be squashed before merging.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "guzzlehttp/ringphp" : "~1.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "^4.7|^5.4",
+    "phpunit/phpunit": "5.7.21",
     "mockery/mockery": "0.9.4",
     "symfony/yaml": "^2.8",
     "symfony/finder": "^2.8",
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "guzzlehttp/ringphp" : "~1.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "5.7.21",
+    "phpunit/phpunit": "6.3.0",
     "mockery/mockery": "0.9.4",
     "symfony/yaml": "^2.8",
     "symfony/finder": "^2.8",
--- a/phpunit-integration.xml
+++ b/phpunit-integration.xml
@@ -1,14 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/5.7/phpunit.xsd"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.3/phpunit.xsd"
          bootstrap="tests/bootstrap.php"
          colors="true"
          failOnRisky="true"
          verbose="true"
-         backupGlobals="false"
          beStrictAboutChangesToGlobalState="true"
          beStrictAboutOutputDuringTests="true"
-         beStrictAboutTestsThatDoNotTestAnything="true"
 >
     <php>
         <env name="ES_TEST_HOST" value="http://localhost:9200"/>
--- a/phpunit-integration.xml
+++ b/phpunit-integration.xml
@@ -1,17 +1,20 @@
-<phpunit
-        bootstrap="tests/bootstrap.php"
-        colors="true"
-        convertNoticesToExceptions="true"
-        convertWarningsToExceptions="true"
-        convertErrorsToExceptions="true"
-        syntaxCheck="true"
-        verbose="true"
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/5.7/phpunit.xsd"
+         bootstrap="tests/bootstrap.php"
+         colors="true"
+         failOnRisky="true"
+         verbose="true"
+         backupGlobals="false"
+         beStrictAboutChangesToGlobalState="true"
+         beStrictAboutOutputDuringTests="true"
+         beStrictAboutTestsThatDoNotTestAnything="true"
 >
     <php>
         <env name="ES_TEST_HOST" value="http://localhost:9200"/>
     </php>
     <testsuites>
-        <testsuite>
+        <testsuite name="Integration Tests">
             <file>tests/Elasticsearch/Tests/YamlRunnerTest.php</file>
         </testsuite>
     </testsuites>
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,17 +1,20 @@
-<phpunit
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/5.7/phpunit.xsd"
         bootstrap="tests/bootstrap.php"
         colors="true"
-        convertNoticesToExceptions="true"
-        convertWarningsToExceptions="true"
-        convertErrorsToExceptions="true"
-        syntaxCheck="true"
+        failOnRisky="true"
         verbose="true"
+        backupGlobals="false"
+        beStrictAboutChangesToGlobalState="true"
+        beStrictAboutOutputDuringTests="true"
+        beStrictAboutTestsThatDoNotTestAnything="true"
 >
     <php>
-        <env name="ES_TEST_HOST" value="http://localhost:9200"/>
+        <server name="ES_TEST_HOST" value="http://localhost:9200"/>
     </php>
     <testsuites>
-        <testsuite>
+        <testsuite name="Tests">
             <directory>tests</directory>
             <exclude>tests/Elasticsearch/Tests/YamlRunnerTest.php</exclude>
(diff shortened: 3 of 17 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
     "guzzlehttp/ringphp" : "~1.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "^4.7|^5.4",
+    "phpunit/phpunit": "6.3.0",
```
mhujer: The `require-dev` does not affect the user, so it is best to use specific version.

phpunit-integration.xml
```diff
@@ -1,17 +1,18 @@
...
-        bootstrap="tests/bootstrap.php"
-        colors="true"
-        convertNoticesToExceptions="true"
-        convertWarningsToExceptions="true"
-        convertErrorsToExceptions="true"
-        syntaxCheck="true"
-        verbose="true"
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/6.3/phpunit.xsd"
```
mhujer: PHPUnit offers XSD schema which allows you to validate the config (and the IDE can use it for "code completion" - PHPStorm does!)

phpunit-integration.xml
```diff
@@ -1,17 +1,18 @@
-<phpunit
-        bootstrap="tests/bootstrap.php"
-        colors="true"
-        convertNoticesToExceptions="true"
```
mhujer: the removed ones either no longer exist or are defaults in new version of PHPUnit

phpunit.xml
```diff
@@ -1,17 +1,18 @@
...
-        convertErrorsToExceptions="true"
-        syntaxCheck="true"
+        failOnRisky="true"
         verbose="true"
+        beStrictAboutChangesToGlobalState="true"
+        beStrictAboutOutputDuringTests="true"
 >
     <php>
-        <env name="ES_TEST_HOST" value="http://localhost:9200"/>
+        <server name="ES_TEST_HOST" value="http://localhost:9200"/>
```
mhujer: It is used as `$_SERVER['ES_TEST_HOST']` in tests, so it should be like this.

tests/Elasticsearch/Tests/ClientTest.php
```diff
@@ -19,18 +20,18 @@
  * @license    http://www.apache.org/licenses/LICENSE-2.0 Apache2
  * @link       http://elasticsearch.org
  */
-class ClientTest extends \PHPUnit_Framework_TestCase
+class ClientTest extends \PHPUnit\Framework\TestCase
```
mhujer: New PHPUnit uses namespaces

tests/Elasticsearch/Tests/ClientTest.php
```diff
@@ -19,18 +20,18 @@
...
     {
         m::close();
     }
 
-    /**
-     * @expectedException \Elasticsearch\Common\Exceptions\InvalidArgumentException
-     */
     public function testConstructorIllegalPort()
     {
+        $this->expectException(\Elasticsearch\Common\Exceptions\InvalidArgumentException::class);
```
mhujer: `$this->expectException()` is much safer that the annotation because of referencing the class directly (so it would fail on PHP level, if the class is removed or renamed)

tests/Elasticsearch/Tests/ClientTest.php
```diff
@@ -49,6 +50,8 @@ public function testCustomQueryParams()
             'client' => ['ignore' => 400]
         ];
         $exists = $client->exists($getParams);
+
+        $this->assertFalse($exists);
```
mhujer: The tests without any assertions are considered "risky" by PHPUnit

tests/Elasticsearch/Tests/ClientTest.php
```diff
@@ -88,136 +94,148 @@ public function testFromConfigBadParamQuiet()
             'imNotReal' => 5
         ];
         $client = ClientBuilder::fromConfig($params, true);
+
+        $this->assertInstanceOf(Client::class, $client);
     }
 
-    public function testNullDelete()
+    public function testIndexCannotBeNullForDelete()
```
mhujer: Here I split the original methods with several try/catch blocks into distinct methods with `$this->expectException` and `$this->expectExceptionMessage`

tests/Elasticsearch/Tests/ClientTest.php
```diff
@@ -88,136 +94,148 @@ public function testFromConfigBadParamQuiet()
...
+            'id' => 'test'
+        ]);
+    }
+
+    public function testIndexCannotBeArrayOfNullsForDelete()
+    {
+        $client = ClientBuilder::create()->build();
+
+        $this->expectException(Elasticsearch\Common\Exceptions\InvalidArgumentException::class);
+        $this->expectExceptionMessage('index cannot be an array of empty strings');
```
mhujer: The message is slightly incorrect, but I don't think it is necessary to change it. (the code does not differ between null and empty string https://github.com/mhujer/elasticsearch-php/blob/132e291fb2e36dad5c6c538ecb6e1cbfa16d2fd8/src/Elasticsearch/Client.php#L1553)

tests/Elasticsearch/Tests/ConnectionPool/SniffingConnectionPoolTest.php
```diff
@@ -311,8 +308,10 @@ public function testAddSeed_SniffTwo_TimeoutTwo()
         ];
         $connectionPool = new SniffingConnectionPool($connections, $selector, $connectionFactory, $connectionPoolParams);
 
+        $this->expectException(\Elasticsearch\Common\Exceptions\NoNodesAvailableException::class);
+        $this->expectExceptionMessage('No alive nodes found in your cluster');
+
         $retConnection = $connectionPool->nextConnection();
-        $this->assertSame($mockConnection, $retConnection);
```
mhujer: it never got here as the previous line throws an exception

tests/Elasticsearch/Tests/ConnectionPool/SniffingConnectionPoolTest.php
```diff
@@ -412,8 +408,6 @@ public function testTen_TimeoutNine_SniffTenth_AddTwoDead_TimeoutEveryone()
                     ->andReturnValues($newConnections)
                     ->getMock();
 
-        $RRConnections = $newConnections;
```
mhujer: unused variable

tests/Elasticsearch/Tests/ConnectionPool/SniffingConnectionPoolTest.php
```diff
@@ -424,10 +418,9 @@ public function testTen_TimeoutNine_SniffTenth_AddTwoDead_TimeoutEveryone()
         ];
         $connectionPool = new SniffingConnectionPool($connections, $selector, $connectionFactory, $connectionPoolParams);
 
-        $retConnection = $connectionPool->nextConnection();
-        $this->assertSame($newConnections[11], $retConnection);
```
mhujer: the diff is a bit confusing here, the first `$retConnection = $connectionPool->nextConnection();` throws an exception, so the rest was a dead code

tests/Elasticsearch/Tests/ConnectionPool/StaticConnectionPoolTest.php
```diff
@@ -224,13 +224,9 @@ public function testCustomConnectionPoolIT()
...
 
-        try {
-            $client->search([]);
-            $this->fail("Should have thrown NoNodesAvailableException");
-        } catch (Elasticsearch\Common\Exceptions\NoNodesAvailableException $e) {
-            // All good
-        } catch (\Exception $e) {
-            throw $e;
-        }
+        $this->expectException(Elasticsearch\Common\Exceptions\NoNodesAvailableException::class);
```
mhujer: PHPUnit checks that the thrown exception is of a proper type

tests/Elasticsearch/Tests/RegisteredNamespaceTest.php
```diff
@@ -34,14 +34,15 @@ public function testRegisteringNamespace()
...
     }
 
-    /**
-     * @expectedException \Elasticsearch\Common\Exceptions\BadMethodCallException
-     */
     public function testNonExistingNamespace()
     {
         $builder = new FooNamespaceBuilder();
         $client = ClientBuilder::create()->registerNamespace($builder)->build();
-        $this->assertSame("123", $client->bar()->fooMethod());
```
mhujer: the assert wasn't used as there was an exception thrown

polyfractal:
<3  I need to attend to some other things, but will look over this tomorrow.  Thanks!

mhujer:
There are conflicts with the recent changes, I will rebase it in the morning (GMT+2)

mhujer:
@polyfractal ping? ;-)

polyfractal:
Heya, sorry, been heads-down on core java work.  I'll try to look over this (and the other PRs) today.

Thanks for the ping! :)

mhujer:
No problem! The Elastic itself is more important than some polishing here, this can wait after the 6.0 release.

polyfractal:
Not necessarily more important... without clients using ES would be much less pleasant!  I'm just bad at multitasking :)

LGTM, thanks again for all the hard work here!

PR #633, opened 2017-08-23
Add possibility of choosing the default URL of project by .env

pedromazala:
The method `Elasticsearch\ClientBuilder::getDefaultHost()` does not have the possibility of choosing our default host of Elastic Search. With this PR, we will can chose by adding the entry `ELASTICSEARCH_DEFAULT_HOST` in our `.env` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,15 @@ Version Matrix
  - You should never use Elasticsearch-PHP Master branch, as it tracks Elasticearch master and may contain incomplete features or breaks in backwards compat.  Only use ES-PHP master if you are developing against ES master for some reason.
 
 Documentation
---------------
+-------------
 [Full documentation can be found here.](http://www.elasticsearch.org/guide/en/elasticsearch/client/php-api/5.0/index.html)  Docs are stored within the repo under /docs/, so if you see a typo or problem, please submit a PR to fix it!
+
+Environment
+-----------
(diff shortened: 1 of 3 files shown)


rafaelpadovani:
+1

rafaelpadovani:
+1

carlosalexandre1985:
It's very useful in Docker (integration between containers).

polyfractal:
I'm not outright opposed to it, but the change does feel a bit more opinionated than I'm normally comfortable with.

Is there a reason this can't be done when the client is being constructed, e.g. by just supplying a host instead of relying on the default?

polyfractal:
Also, it appears to break the client if the dotfile doesn't exist at all: https://travis-ci.org/elastic/elasticsearch-php/builds/267547153 :)

pedromazala:
That is true 😞 
But I can fix it... The reason is because some packages that use this package does not given a default value to the host and it breaks...

pedromazala:
Actually, looking the code, they have a way to set the `ClientBuilder` config using the method `formConfig` 🙈 . But they dont show it to us in the docs. 
Thank you for your time :)

polyfractal:
Sure thing :)  Let me know if you want me to comment on an issue with the downstream package, I'd be happy to help the maintainer with the integration.

PR #645, opened 2017-09-09
update Simulate API as it uses POST instead of GET

p365labs:
Ingest Simulate Endpoint has GET as HTTP method and instead it should be a **POST**

https://www.elastic.co/guide/en/elasticsearch/reference/current/simulate-pipeline-api.html

### DIFF
--- a/src/Elasticsearch/Endpoints/Ingest/Simulate.php
+++ b/src/Elasticsearch/Endpoints/Ingest/Simulate.php
@@ -60,6 +60,6 @@ class Simulate extends AbstractEndpoint
      */
     public function getMethod()
     {
-        return 'GET';
+        return 'POST';
     }
 }


polyfractal:
The Simulate endpoint accepts both GET and POST: https://github.com/elastic/elasticsearch/blob/master/rest-api-spec/src/main/resources/rest-api-spec/api/ingest.simulate.json#L4

Did you run into an error, or were just changing to match the docs?

p365labs:
@polyfractal No, I didn't run into any errors... I only read the [documentation](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/simulate-pipeline-api.html) but I did not have a look to the rest-api-spec you reported. Maybe the docs should be updated, is not really clear. I think we can close this.

polyfractal:
:+1:

PR #647, opened 2017-09-14
[src] add PHP_CodeSniffer

mhujer:
This PR fixes PSR-2 violations in `src` and adds automatic check as a part of the build.

**Depends on #638 (it should be merged first)**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,6 +44,7 @@
   },
   "scripts": {
     "phpcs": [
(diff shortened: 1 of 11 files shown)


polyfractal:
Looks like this is the next PR in the chain of dependencies?

mhujer:
@polyfractal Just rebased it.

polyfractal:
emot-toot

PR #658, opened 2017-10-08
Integration test with UTF-8 char

Nexucis:
Hello,

I'm coming with the tests as we discussed in the PR #482. The problem is only when we create an index and an alias with some UTF-8 characters and after that we try to get the index by the alias or we want to verify if the given alias exists.

As a workaround, we can urlencode the alias before passed it to the client : 

```php
$params = array(
            'name' => urlencode($alias)
);
// return true even with UTF-8 char
$this->assertEquals($alias . '_v1', array_keys(self::$client->indices()->getAlias($params))[0]);
```

I run the following test with ElasticSearch 2.4 and 5.5.2

### DIFF
--- a/tests/Elasticsearch/Tests/ClientIntegrationTests.php
+++ b/tests/Elasticsearch/Tests/ClientIntegrationTests.php
@@ -76,8 +76,8 @@ class ClientIntegrationTests extends \PHPUnit\Framework\TestCase
     public function existAliasDataProvider()
     {
         return [
-            ['myindextest'],
-            ['⿇⽸⾽']
+            'latin-char' => ['myindextest'],
+            'utf-8-char' => ['⿇⽸⾽']
         ];
     }
 
--- a/tests/Elasticsearch/Tests/ClientIntegrationTests.php
+++ b/tests/Elasticsearch/Tests/ClientIntegrationTests.php
@@ -102,7 +102,7 @@ class ClientIntegrationTests extends \PHPUnit\Framework\TestCase
         self::$client->indices()->putAlias($params);
 
         $params = array(
-            'name' => urlencode($alias)
+            'name' => $alias
         );
         $this->assertEquals($alias . '_v1', array_keys(self::$client->indices()->getAlias($params))[0]);
     }
--- a/tests/Elasticsearch/Tests/ClientIntegrationTests.php
+++ b/tests/Elasticsearch/Tests/ClientIntegrationTests.php
@@ -1,10 +1,11 @@
 <?php
 
-declare(strict_types = 1);
+declare(strict_types=1);
 
 namespace Elasticsearch\Tests;
 
 use Elasticsearch;
+use stdClass;
 
 /**
  * Class ClientTest
@@ -33,6 +34,47 @@ class ClientIntegrationTests extends \PHPUnit\Framework\TestCase
         $exists = $client->exists($getParams);
 
         $this->assertFalse($exists);
+    }
+
+    /**
+     * @dataProvider
+     */
+    public function testExistAlias($alias)
+    {
+        $client = Elasticsearch\ClientBuilder::create()->setHosts([$_SERVER['ES_TEST_HOST']])->build();
+
+        $this->createIndexWithAlias($alias, $client);
+
+        $params = array(
+            'name' => $alias
+        );
+
+        $this->assertTrue($client->indices()->existsAlias($params));
+    }
+
+    public function existAliasDataProvider()
+    {
+        return [
+            ['myindextest'],
+            ['⿇⽸⾽']
+        ];
+    }
+
+    /**
+     * @param $alias
+     * @param $client Elasticsearch\Client
+     */
+    private function createIndexWithAlias($alias, $client)
+    {
+        $params = array(
+            'index' => $alias . '_v1',
+            'body' => array(
+                'aliases' => array(
+                    $alias => new stdClass()
+                )),
+        );
+
+        $client->indices()->create($params);
     }
 
 }
--- a/tests/Elasticsearch/Tests/ClientIntegrationTests.php
+++ b/tests/Elasticsearch/Tests/ClientIntegrationTests.php
@@ -41,8 +41,6 @@ class ClientIntegrationTests extends \PHPUnit\Framework\TestCase
 
     public function testCustomQueryParams()
     {
-        $this->client = Elasticsearch\ClientBuilder::create()->setHosts([$_SERVER['ES_TEST_HOST']])->build();
-
         $getParams = [
             'index' => 'test',
             'type' => 'test',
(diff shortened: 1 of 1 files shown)


Nexucis:
In fact, with the new test `testPutAlias`, we can see that the problem is with the parameter `name`. The parameter `index` seems to work properly :)

Nexucis:
ping @polyfractal :)

tests/Elasticsearch/Tests/ClientIntegrationTests.php
```diff
@@ -1,10 +1,11 @@
 <?php
 
-declare(strict_types = 1);
+declare(strict_types=1);
```
mhujer: should not be changed in the PR

tests/Elasticsearch/Tests/ClientIntegrationTests.php
```diff
@@ -18,9 +19,32 @@
...
+        $param = [
+            'index' => '_all'
+        ];
+        self::$client->indices()->delete($param);
+    }
+
     public function testCustomQueryParams()
     {
-        $client = Elasticsearch\ClientBuilder::create()->setHosts([$_SERVER['ES_TEST_HOST']])->build();
+        self::$client = Elasticsearch\ClientBuilder::create()->setHosts([$_SERVER['ES_TEST_HOST']])->build();
```
mhujer: Is this change necessary? I think that it is better if the tests do not depend on static attribute.

tests/Elasticsearch/Tests/ClientIntegrationTests.php
```diff
@@ -30,8 +54,85 @@ public function testCustomQueryParams()
...
+     * @dataProvider aliasDataProvider
+     */
+    public function testFindIndexByAlias($alias)
+    {
+        $this->createIndexWithAlias($alias);
+
+        $params = array(
+            'name' => $alias
+        );
+        $this->assertEquals($alias . '_v1', array_keys(self::$client->indices()->getAlias($params))[0]);
```
mhujer: `$this->assertSame()` should be used insted of `assertEquals` (there are more occurences in the PR)

`assertSame` is like `===`
`assertEquals` is like `==`

tests/Elasticsearch/Tests/ClientIntegrationTests.php
```diff
@@ -30,8 +54,85 @@ public function testCustomQueryParams()
...
+    }
+
+    /**
+     * @dataProvider aliasDataProvider
+     */
+    public function testPutAlias($alias)
+    {
+
```
mhujer: unnecessary newline

tests/Elasticsearch/Tests/ClientIntegrationTests.php
```diff
@@ -30,8 +54,85 @@ public function testCustomQueryParams()
...
+    public function aliasDataProvider()
+    {
+        return [
+            'latin-char' => ['myindextest'],
+            'utf-8-char' => ['⿇⽸⾽']
+        ];
+    }
+
+    private function createIndexWithAlias($alias)
```
mhujer: there can be a `string` typehint

tests/Elasticsearch/Tests/ClientIntegrationTests.php
```diff
@@ -30,8 +54,85 @@ public function testCustomQueryParams()
...
+            'body' => array(
+                'aliases' => array(
+                    $alias => new stdClass()
+                )),
+        );
+
+        self::$client->indices()->create($params);
+    }
+
+    private function createIndexWithoutAlias($alias)
```
mhujer: there can be a `string` typehint

tests/Elasticsearch/Tests/ClientIntegrationTests.php
```diff
@@ -18,9 +19,32 @@
  */
 class ClientIntegrationTests extends \PHPUnit\Framework\TestCase
 {
+
+    /**
+     * @var $client Elasticsearch\Client
+     */
+    protected static $client;
```
mhujer: What about making it non-static a recreating it for each test in `setUp()`? (it probably does not matter here, but I think it is a good practice to keep tests independent)

Nexucis: well it's just to not initialize the client in every test below. I'm not sure to understand why it makes the test dependent. If you don't want to use the global client, you can ever initialize a new one.

tests/Elasticsearch/Tests/ClientIntegrationTests.php
Nexucis: thanks for the tips, i didn't know that :)

tests/Elasticsearch/Tests/ClientIntegrationTests.php
Nexucis: i'm sorry, i don't understand what do you mean by that :s

mhujer: You can change the method signature to `private function createIndexWithAlias(string $alias)` (added `string`), so the parameter type is checked.

tests/Elasticsearch/Tests/ClientIntegrationTests.php
Nexucis: oh sorry, I read your comment in a wrong way. Well yes we can initialiaze it in a `setUp()` method

tests/Elasticsearch/Tests/ClientIntegrationTests.php
Nexucis: thanks for the explanation :)

Nexucis:
These new commit should fixe the review made by @mhujer (thanks for the review by the way)

tests/Elasticsearch/Tests/ClientIntegrationTests.php
```diff
@@ -18,9 +19,29 @@
...
+        $param = [
+            'index' => '_all'
+        ];
+        $this->client->indices()->delete($param);
+    }
+
     public function testCustomQueryParams()
     {
-        $client = Elasticsearch\ClientBuilder::create()->setHosts([$_SERVER['ES_TEST_HOST']])->build();
+        $this->client = Elasticsearch\ClientBuilder::create()->setHosts([$_SERVER['ES_TEST_HOST']])->build();
```
mhujer: I think this line can be removed completely, because `$this->client` is already initialized in `setUp()`

Nexucis: luckily you're here, I completely forgot to remove it. Thanks again

Nexucis:
ping @polyfractal :)

Nexucis:
pong @polyfractal 😏

Nexucis:
For an undkown reason, the test don't pass with `docker.elastic.co/elasticsearch/elasticsearch:6.0.0`. Which seems different here, because all test pass. 

So I don't  if there is a problem with the docker image or with the current ci here.

Maybe it's another reason to consider the issue #693 (i'm kidding)

Nexucis:
ping @polyfractal and Happy new Years :)

ezimuel:
Closing because it's related with #482.

Issue #659, opened 2017-10-10
Unable to set 'verify' to 'false' while creating repository


ParitoshBh:
### Summary of problem or feature request

I am trying to use the `snapshot()->createRepository()` end-point for creating a new repository in a cluster with 2 nodes. As per the [Snapshot/Restore documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html#_repository_verification) setting `verify` to `false` would disable the repository verification.

However when I try to use the same with the library as following,

```
    $params = [
      'repository' => 'repo_test',
      'body' => [
        'type' => 'fs',
        'verify' => false,
        'settings' => [
          'compress' => true,
          'location' => '/usr/local/elasticsearch/backups'
        ]
      ]
    ];
```

The repository verification is still carried out as if the `verify` param isn't there at all! Tried using `verify` as part of `settings` param gave the same results as well,

```
        'settings' => [
          'compress' => true,
          'location' => '/usr/local/elasticsearch/backups',
          'verify' => false
        ]
```

Not really sure if this is a bug or a general support question.

### System details

- Operating System - Cent OS 7
- PHP Version - PHP 7.1.10
- ES-PHP client version - v5.3.0
- Elasticsearch version - v5.6.1

polyfractal:
Sorry for the delay, was on holiday and then a week-long work trip.

The `verify` setting is a query string parameter, meaning it needs to go into the top-level of the request params rather than the body.  Try this:

```php
   $params = [
      'repository' => 'repo_test',
      'verify' => false,  // <-- here
      'body' => [
        'type' => 'fs',
        'settings' => [
          'compress' => true,
          'location' => '/usr/local/elasticsearch/backups'
        ]
      ]
    ];
```

ParitoshBh:
Did try that as well but doing so results in following error message,

```
Exception: "verify" is not a valid parameter. Allowed parameters are "client", 
"custom", "filter_path", "human", "master_timeout", "timeout" in 
[/var/www/elasticsearch/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Endpoints
/AbstractEndpoint.php, line 237]
```

Snippet I am using,

```php
    $params = [
      'repository' => 'test_repo',
      'verify' => false,
      'body' => [
        'type' => 'fs',
        'settings' => [
          'compress' => true,
          'location' => '/usr/local/elasticsearch/backups'
        ]
      ]
    ];
```

[...]

PR #665, opened 2017-10-24
Add verify as acceptable query string parameter for create repository

ParitoshBh:
Addresses missing `verify` parameter issue in #659


polyfractal:
Thanks!  Sorry for the delay :(

PR #675, opened 2017-11-07
PHPStan check src/ on level 2

mhujer:
I've fixed all issues found by PHPStan in `src/` on level 2 and added it as a part of the build.

**This depends on #674**


polyfractal:
As an aside, while it drives me crazy that PHP methods/functions are case-insensitive... it does mean these aren't breaking changes which is quite nice :)

PR #682, opened 2017-11-15
Handle ServerErrorResponseException on promise failure

chs2:
\Elasticsearch\Connections\Connection::process5xxError() throws exception which is passed to rejected promise as argument.
Then $response is no longer an array. The previous test triggered "Error: Cannot use object of type Elasticsearch\Common\Exceptions\ServerErrorResponseException as array".

### DIFF
--- a/src/Elasticsearch/Transport.php
+++ b/src/Elasticsearch/Transport.php
@@ -119,7 +119,8 @@ class Transport
             //onFailure
             function ($response) {
                 // Ignore 400 level errors, as that means the server responded just fine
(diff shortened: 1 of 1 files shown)


polyfractal:
This is/was a bug when using the client with PHP 5.  It's been fixed in https://github.com/elastic/elasticsearch-php/commit/c5db567d5a843a2c9778c7b74863357879abd34f for the ES-PHP 5.x branch.

This works fine on PHP7 however, which is what master and 6.x branch require (see: https://github.com/elastic/elasticsearch-php#php-version-requirement)

Thanks though!

PR #691, opened 2017-12-03
[X-PACK / PART 1] Info API

Nexucis:
Hello,

I'm coming with an implementation of the xpack api.

This PL concern only the [Info API](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/info-api.html). The other api will come in others PL.

The X-pack API is available since elasticsearch 5.5, I think, so if your are interested, I can backport this PL in the 5.0 branche.


Nexucis:
Well something don't work with my new integration test. I don't get it. It works when I ran it with IntelliJ with phpunit.xml as configuration files. But with travis, all of them failed :(

mhujer:
I guess that's because XPack is not available in the Elasticsearch which is used for Travis.

Nexucis:
Well, i tried with a docker image in 6.0.0 version and the version used by travis is 6.0.1-SNAPSHOT. So XPack should be available in both version, isn't it ?

Nexucis:
Fun facts, if I use `docker.elastic.co/elasticsearch/elasticsearch-platinum:6.0.0` instead of `docker.elastic.co/elasticsearch/elasticsearch:6.0.0`, the tests failed, because of a http error `401`

But I think it's normal because in a trial cluster, x-pack is available but not enable. While in a platinum version, x-pack is available AND enable.

polyfractal:
Heya Nexucis, we have an XPack plugin here:  https://github.com/elastic/elasticsearch-x-pack-php

It's separate from the main ES-PHP repo because we didn't want to put proprietary APIs into the OSS client.  E.g. it would be mean to developers to see APIs they can't access because they don't have XPack installed.  Also because of things like Travis testing.

The XPack plugin should be mostly up-to-date, although I need to do a post-6.0 check and make sure.

Nexucis:
Oh ok , I didn't see that.

Well this PL is no longer useful. I'll close this PL and #692 

Thanks for the infos @polyfractal .

PR #692, opened 2017-12-03
[X-PACK / PART 2] User API

Nexucis:
And here we go, 2nd PL for X-pack API. 

You can find all endpoint describe in the [user api](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/security-api-users.html#security-api-users)

This PL must be merged after #691 

Sadly, i don't have a gold license to write some integration test :( . Sorry about that

And of course I can backport this PL in the 5.0 branches if you want


Nexucis:
Ok, I found the docker image `docker.elastic.co/elasticsearch/elasticsearch-platinum:6.0.0` which give a trial platinium license. So this should be enough to create some integration test. 

But, maybe I need to isolate this kind of test, in order to not disturb the launch of the other tests that do not need a particular license. For example with a new phpunit configuration file and if the issue #693 is accepted, then with a new job. Thoughts ?

Nexucis:
A X-pack plugin already exists. So this code is useless. See #691 for more information

PR #694, opened 2017-12-04
Add note about separate X-Pack library to README

mhujer:
As mentioned in https://github.com/elastic/elasticsearch-php/pull/691#issuecomment-348980615, there is a [standalone XPack plugin](https://github.com/elastic/elasticsearch-x-pack-php). It should be written down in the README, so the users can discover it easily.

(I also fixed some README typos in the first commit)


polyfractal:
You'd think I'd be able to spell Elasticsearch after all this time :)

:+1:

Issue #702, opened 2017-12-08
`search_after` parameter not whitelisted/usable


rtzm:
### Summary of problem or feature request

I am trying to use the `search_after` request body parameter (https://www.elastic.co/guide/en/elasticsearch/reference/5.4/search-request-search-after.html), but I'm receiving errors that this is not a whitelisted parameter on the Search endpoint.

### Code snippet of problem

I'm using this basemkhirat/elasticsearch library in my code, and have opened a pull request to allow the `search_after` parameter. (This is where that library interacts with the `Elasticsearch\Client->search()` method: https://github.com/basemkhirat/elasticsearch/pull/63/files#diff-4528a13e17ddcea8ae7919377e58843bR895.) 

However, when I try to use this code, I'm getting this error:
```
Elasticsearch\Common\Exceptions\UnexpectedValueException
"search_after" is not a valid parameter.  Allowed parameters are "_source", "_source_exclude", "_source_include", "allow_no_indices", "analyze_wildcard", "analyzer", "batched_reduce_size", "client", "custom", "default_operator", "df", "docvalue_fields", "expand_wildcards", "explain", "fielddata_fields", "filter_path", "filter_path", "from", "human", "ignore_unavailable", "indices_boost", "lenient", "lowercase_expanded_terms", "preference", "q", "query_cache", "request_cache", "routing", "scroll", "search_type", "size", "slice", "sort", "source", "stats", "stored_fields", "suggest_field", "suggest_mode", "suggest_size", "suggest_text", "terminate_after", "timeout", "typed_keys", "version"

in /srv/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Endpoints/AbstractEndpoint.php line 237

at Elasticsearch\Endpoints\AbstractEndpoint->checkUserParams(array('from' => 0, 'size' => 2, 'search_after' => array('1512483884000', 'contact#AWAnEKRrboRQijOz8agk'))) in /srv/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Endpoints/AbstractEndpoint.php line 74

at Elasticsearch\Endpoints\AbstractEndpoint->setParams(array('from' => 0, 'size' => 2, 'search_after' => array('1512483884000', 'contact#AWAnEKRrboRQijOz8agk'))) in /srv/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Client.php line 950

[...]

Issue #556, opened 2017-03-19
What about "Search after" feature?


KiPa187:
### Summary of problem or feature request

[Search after](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-search-after.html) is not implemented in PHP client!
And how much more features was not released in "version 5" of PHP client?
Please, fix it asap.

### System details

- Operating System
- PHP Version
- ES-PHP client version
- Elasticsearch version

KiPa187:
OK, it's works if I include "search_after" param in "body" param..
But it's not clear and not documented.

cviebrock:
Submit a PR with updated documentation, please.

PR #703, opened 2017-12-08
Added note in doc block regarding non-whitelisted params for search()

rtzm:
Currently, the params for `Client->search()` are whitelisted, and that whitelisting doesn't clearly match the expectations laid out in the Search APIs documentation for Elasticsearch (e.g. https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-request-search-after.html). This PR adds some documentation explaining usage for some of those params which are not documented elsewhere, and which throw errors which make it seem that those params are not supported by this library.

Related issues:
https://github.com/elastic/elasticsearch-php/issues/702
https://github.com/elastic/elasticsearch-php/issues/556

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -929,6 +929,10 @@ class Client
      *        ['version']                  = (boolean) Specify whether to return document version as part of a hit
      *        ['body']                     = (array|string) The search definition using the Query DSL
      *
+     * Note: Some Search API parameters, such as `search_after`, are not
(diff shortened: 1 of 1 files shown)


polyfractal:
I'm not sure I understand.  `search_after` is clearly a component of the body, and the ES reference documentation doesn't say it can be used as part of the URL query string.

E.g. `aggs`, `highlight`, `suggest` etc are not available in the query string either.  I don't think we need a disclaimer for such things?

PR #711, opened 2017-12-28
[travis] use very specific ES version for build

mhujer:
This changes the Travis build, so it uses ES releases for builds instead of snapshots.

It started to fail because for some reason, commit from which the snapshot was build, was not available in the ES repository anymore ([See the Travis log](https://travis-ci.org/mhujer/elasticsearch-php/jobs/322340609))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,20 +18,20 @@ matrix:
   fast_finish: true
   include:
     - php: 7.0
-      env: ES_VERSION="6.0"
+      env: ES_VERSION="6.0.1"
     - php: 7.0
-      env: ES_VERSION="6.x"
+      env: ES_VERSION="6.1.1"
 
     - php: 7.1
-      env: ES_VERSION="6.0"
+      env: ES_VERSION="6.0.1"
     - php: 7.1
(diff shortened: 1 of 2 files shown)


mhujer:
@polyfractal ping? ;-) This one is required for any other PR build to pass...

polyfractal:
Heya, just got back from holiday :)

So after some investigation, this seems to be an issue with how Kibana's `esvm` is generating snapshots.  It appears to be picking up the commit from the snapshot builder itself (via an environment variable that Jenkins grabs which then ES grabs) which is why it doesn't exist in the ES git tree.

I've poked the relevant people, will update as more info comes up.

Would prefer to keep these pointing at snapshots rather than hard versions, so we don't have to bump them all the time.  Will keep you apprised as more details emerge :)

polyfractal:
And here we go: https://github.com/elastic/elasticsearch/pull/28082/files

Life should be back to normal after that's merged and backported

mhujer:
Cool 👍

mhujer:
Thanks for having it fixed in upstream! I've rebased by my other PRs directly on master.

PR #717, opened 2018-01-05
[TravisCI] update build matrix

mhujer:
I've updated build matrix to also build against ES 6.1 (it was only using 6.0 and 6.2-dev before).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,15 +20,19 @@ matrix:
     - php: 7.0
       env: ES_VERSION="6.0"
     - php: 7.0
-      env: ES_VERSION="6.x"
+      env: ES_VERSION="6.1"
 
     - php: 7.1
       env: ES_VERSION="6.0"
     - php: 7.1
-      env: ES_VERSION="6.x"
+      env: ES_VERSION="6.1"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,15 +20,19 @@ matrix:
...
     - php: 7.1
-      env: ES_VERSION="6.x"
+      env: ES_VERSION="6.1"
 
     - php: 7.2
       env: ES_VERSION="6.0"
+    - php: 7.2
+      env: ES_VERSION="6.1"
+    - php: 7.2
+      env: ES_VERSION="6.x"
```
mhujer: it makes sense to run only one build against unreleased 6.x version (it will be probably failing anyway)

mhujer:
Surprisingly, some tests are failing against ES 6.1.2.

polyfractal:
Thanks @mhujer :)  Finally getting some time to put towards the client.

Merging this now, will update it shortly with the current list of releases since this is a bit out of date now (due to my negligence) :)

mhujer:
@polyfractal 👍  :-)

Issue #718, opened 2018-01-06
Cannot fetch status of a snapshot within a repository


sarcastron:
### Summary of problem or feature request

I believe I have found a bug with the snapshot namespace. When attempting to get information about a snapshot within a repository, only the status data regarding the repository is returned.

The bug exists in the `Elasticsearch\Endpoints\Snapshot\Status::getURI()` method.

As far as I can tell this issue exists in version 5 and 6.

### Code snippet of problem

After creating a repository of `my_backup` and a snapshot called `snapshot_1`
```php
    $result = $esClient->snapshot()->status([
        'repository' => 'my_backup',
        'snapshot' => 'snapshot_1',
    ]);

    var_export($result);
```

Response is:
```php
[
  "snapshots" => []
]
```

Response should be something like:
```php
"snapshots" => array:1 [
    0 => array:7 [
      "snapshot" => "snapshot_1"
      "repository" => "my_backup"
      "uuid" => "D3rp3AcLTMqYv0E5coyNyw"
      "state" => "SUCCESS"
      "shards_stats" => array:6 [
        "initializing" => 0
        "started" => 0
        "finalizing" => 0
        "done" => 12
        "failed" => 0
        "total" => 12
      ]
      "stats" => array:6 [
        "number_of_files" => 102
        "processed_files" => 102
        "total_size_in_bytes" => 603627
        "processed_size_in_bytes" => 603627
        "start_time_in_millis" => 1515180506544
        "time_in_millis" => 1093
      ]
      "indices" => array:1 [
        "test-blog-posts" => array:3 [
          "shards_stats" => array:6 [
            "initializing" => 0
            "started" => 0
            "finalizing" => 0
            "done" => 2
            "failed" => 0
            "total" => 2
          ]
          "stats" => array:6 [
            "number_of_files" => 17
            "processed_files" => 17
            "total_size_in_bytes" => 249637
            "processed_size_in_bytes" => 249637
            "start_time_in_millis" => 1515180506544
            "time_in_millis" => 388
          ]
          "shards" => array:2 [
            0 => array:2 [
              "stage" => "DONE"
              "stats" => array:6 [
                "number_of_files" => 10
                "processed_files" => 10
                "total_size_in_bytes" => 155361
                "processed_size_in_bytes" => 155361
                "start_time_in_millis" => 1515180506709
                "time_in_millis" => 223
              ]
            ]
            1 => array:2 [
              "stage" => "DONE"
              "stats" => array:6 [
[...]

PR #719, opened 2018-01-06
Fix issue with getting status of repository and snapshots.

sarcastron:
This PR hopefully closes [issues 718](https://github.com/elastic/elasticsearch-php/issues/718).

The way the conditions were ordered, paired with if check above, the `elseif` condition could never occur. This PR flips the `if...elseif` logic around so both conditions exist in the correct circumstances.

I tested this under both conditions and got the expected results.

Thank you.

### DIFF
--- a/src/Elasticsearch/Endpoints/Snapshot/Status.php
+++ b/src/Elasticsearch/Endpoints/Snapshot/Status.php
@@ -70,10 +70,10 @@ class Status extends AbstractEndpoint
         $snapshot   = $this->snapshot;
         $uri        = "/_snapshot/_status";
 
-        if (isset($repository) === true) {
-            $uri = "/_snapshot/$repository/_status";
-        } elseif (isset($repository) === true && isset($snapshot) === true) {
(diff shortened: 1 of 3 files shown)


sarcastron:
The changes to `SniffingConnectionPoolIntegrationTest.php` were made because that test was failing and I noticed that other tests involving connection pool sniffing were skipped.

I don't understand enough about that aspect of the unit tests. If this modification is in error, I will revert the file.

sarcastron:
@polyfractal Just checking in on this PR. Is there anything more I should add or change?

polyfractal:
Hi @sarcastron, I'm so sorry for the radio silence.  I've been swamped with another project that I'm trying to finish up, and Elasticon is just around the corner too.

I think this LGTM, thanks for the PR and putting up with my negligence.  I promise it doesn't usually take this long!  Not the experience I usually try to give to new contributors to the project :(

sarcastron:
No worries @polyfractal Thanks for taking the time to review it.

PR #730, opened 2018-02-14
Fix documentation example of upsert

zakirovir6:
Body param of upsert should be like described in ES Update API


polyfractal:
Thanks!  Sorry for the delay :)

PR #736, opened 2018-03-07
[DOCS] Replace deprecated string type with keyword type for index ops

hacfi:


polyfractal:
:+1: Thanks!  Sorry for the delay on this

PR #663, opened 2017-10-20
fix for invalid GET /_aliases route.

Artistan:
_aliases is no longer valid.
https://github.com/elastic/elasticsearch/search?utf8=%E2%9C%93&q=%2F_alias&type=

PR #740, opened 2018-03-20
Add listIndices to Indices namespace

JoffreyPoreeCoding:
Hi,
My app need to list aliases so I add this function.

Thanks you!

### DIFF
--- a/src/Elasticsearch/Endpoints/Indices/Aliases/Get.php
+++ b/src/Elasticsearch/Endpoints/Indices/Aliases/Get.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Elasticsearch\Endpoints\Indices\Aliases;
+
+use Elasticsearch\Endpoints\AbstractEndpoint;
+use Elasticsearch\Common\Exceptions;
+
+/**
+ * Class Get
+ *
+ * @category Elasticsearch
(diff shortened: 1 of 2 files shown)


JoffreyPoreeCoding:
Hello,
I've signed the CLA, and valid email verification.

polyfractal:
Hi @JoffreyPoreeCoding, sorry for the delay.  What version of Elasticsearch are you using?

We [removed the GetAliases endpoint](https://github.com/elastic/elasticsearch-php/pull/663) recently, because it was deprecated and then [removed in Elasticsearch 5.x](https://github.com/elastic/elasticsearch/pull/13906).  Get/Put/Exists/Delete Alias endpoints are all available in the [Alias namespace](https://github.com/elastic/elasticsearch-php/tree/master/src/Elasticsearch/Endpoints/Indices/Alias) (note this is different from the [Aliases namespace](https://github.com/elastic/elasticsearch-php/tree/master/src/Elasticsearch/Endpoints/Indices/Aliases))

If you're using an older version of Elasticsearch (e.g. 2.x) you should be using the 2.0 branch of ES-PHP which [still has the endpoint](https://github.com/elastic/elasticsearch-php/blob/2.0/src/Elasticsearch/Endpoints/Indices/Aliases/Get.php#L16).  

Thanks for the PR though! :)

PR #748, opened 2018-04-04
fix DeleteByQuery param white list

Rabotyahoff:
method delete_by_query can use the q parameter in the same way as the search api


polyfractal:
So it does.  Thanks!  Will forward-port this to 6.x and master as well

PR #759, opened 2018-05-09
Fix Snapshot::getURI() implementation

thePanz:
- [x] the $uri is always returned, plus the repository and snapshot values are double checked
- [x] the isset() returns `true` even when the value is `null` or an empty string
- [x] avoid not setting $repository or $snapshot if the provided value is empty. It could cause errors if the endpoint class is reused

### DIFF
--- a/src/Elasticsearch/Endpoints/Snapshot/AbstractSnapshotEndpoint.php
+++ b/src/Elasticsearch/Endpoints/Snapshot/AbstractSnapshotEndpoint.php
@@ -1,0 +1,121 @@
+<?php
+declare(strict_types = 1);
+
+namespace Elasticsearch\Endpoints\Snapshot;
+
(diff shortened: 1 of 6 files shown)


ezimuel:
@thePanz can you elaborate more about this PR? It looks to me that we need to fix the issues that you pointed out instead of creating a new Abstract class. Moreover, we need to write unit test to proof it need.

ezimuel:
@thePanz I'm closing this PR. I'll reopen it when you will address my previous questions, thanks!

thePanz:
@ezimuel sorry for the delay. The PR started as a fix, and later on I realized that the "issue" is spread around all `Snapshot` endpoints.
I'll provide a better PR with tests, maybe calling it "refactoring" instead of "fix" :)

PR #764, opened 2018-05-18
Few phpdoc block fix

shahariaazam:
Lots of method were supposed to throw exceptions when dealing with app to ELS api. It should be helpful for developer to handle those exceptions to avoid crash.

And also few return type was misleading. I just faced that while I was working with a live project. So I think it's worth sharing with the community.


ezimuel:
@shahariaazam this PR seems to be outdated. It looks like the phpdocs have already been fixed. Let me know if you want to reopen it, thanks!

PR #765, opened 2018-05-18
phpdoc block fix. Added @throw Exception docbloc to avoid API call cr…

shahariaazam:
phpdoc block fix. Added @throw Exception docbloc to avoid API call crash and encourage developer to handle exceptions


ezimuel:
@shahariaazam I'm closing this PR for the same reason of #764. Let me know if this is still relevant, thanks!

PR #774, opened 2018-06-12
fix basic auth for ringPHP

StefanNeuser:
Fixing the HTTP Basic Auth issue with ringPHP.

### DIFF
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -117,8 +117,11 @@ class Connection implements ConnectionInterface
         }
 
         if (isset($hostDetails['user']) && isset($hostDetails['pass'])) {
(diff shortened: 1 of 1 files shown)


polyfractal:
I'm not sure I understand the issue?  The `['client']['curl']` map is used to directly set curl options in RingPHP.  You can see [where that happens here](https://github.com/guzzle/RingPHP/blob/master/src/Client/CurlFactory.php#L40).

So when we set `CURLOPT_HTTPAUTH` and `CURLOPT_USERPWD` in that location, those curlOpts are directly configured into the curl handler by RingPHP.

StefanNeuser:
my bad, i copied the url and got some url encoded signs in it.

polyfractal:
:+1:

PR #778, opened 2018-06-30
add reference to parse_url() for Extended Host Configuration

SirkoS:
Reference the list of possible parameters of the extended host configuration in the docs.


ezimuel:
@SirkoS thanks for your contribution!

Issue #548, opened 2017-02-23
[Header] Making the port optional in the header "Host"


fullmoonissue:
Hi everyone,

In my current use case, I have a nginx above the access to elasticsearch.

The host header created during the process of the request contains the port ("Host: 'website.com:80'").

This actually causes 400 Bad Request or 404 Not Found messages from nginx.

Without any port on the host header, there's no more problem.

Could we add the port into the host header in an configurable way ?

In addition, the [RFC](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html) (Section 14.23) show that the port is optional.

Here is the [line](https://github.com/elastic/elasticsearch-php/blob/master/src/Elasticsearch/Connections/Connection.php#L123) which causes my issue.

What do you think of it ?

Regards,

polyfractal:
Hm, why is your proxy returning 400/404 because of the port?  Is nginx configured to accept connections on port 80, or port 9200?

The RFC says an omitted port implies the "default port for the service requested (e.g., "80" for an HTTP URL)".  In this case, the service's default port is 9200, which is at-odds with the convention of port 80 being the default.  It isn't a strictly HTTP/web server... it's a search engine that exposes it's api over HTTP.  It makes the situation somewhat confusing :)

I'd prefer we see if the proxy configuration can be fixed.  The port may be optional, but the presence of the port should always be handled correctly regardless.  It would be preferable if the client remains simple in this regard, rather than doing somewhat unexpected things like dropping ports.

FullMoonIssue:
Hi @polyfractal,

To complete the picture about how the request is handled, here are the steps :
=> HaProxy receive the request on the :80 port
=> These request is forwarded to Varnish on the :8080 port
=> Then, forward to Nginx on the :82 port
=> Nginx' upstream call ES on the :9200 port.

On the nginx' configuration, it listen on the *:82 port and the server_name does not contain any wilcard after the hostname.

On the one hand, the default port in this case would be correct (port 80), on the other hand the port 9200 would be called but thanks to the different "forwards".

A WIP is still in research to not being affected by the port in the header host (like a wildcard on the server_name or a direct communication between haproxy and ES).

I hope it gives you more details about this particular need and the will of configurable drop of port in the host header.

Regards,

FullMoonIssue:
Hi @polyfractal,

Is there a possibility that this old [PR](https://github.com/elastic/elasticsearch-php/pull/309) could be merged ?
[...]

PR #782, opened 2018-07-13
Ports should be specified in CURLOPT_PORT, not in the host

afrozenpeach:
Because the host is adding the port, the host header also has the port in it. Instead, the handler should be using CURLOPT_PORT to specify which port is the curl request is being sent on. 

This also affects unit tests that no longer need to verify that the port is in the host name.

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -638,7 +638,7 @@ class ClientBuilder
             $host['scheme'] = 'http';
         }
         if (isset($host['port']) === false) {
-            $host['port'] = '9200';
+            $host['port'] = 9200;
         }
         return $host;
     }
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -519,9 +519,6 @@ class Connection implements ConnectionInterface
     {
         return $this->host;
     }
-    
-    /**
-    
 
     /**
      * @return null|string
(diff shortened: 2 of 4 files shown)


polyfractal:
On the run at the moment so haven't looked at this closely... but does it matter?  My understanding is that regardless of where the port is specified (in the URL or in the curl options) the end result is the same?

polyfractal:
Ahh I see your comment in https://github.com/elastic/elasticsearch-php/issues/548#issuecomment-404859017.  Makes sense.  Will try to take a closer look at this later today, or on monday at the latest.

Thanks for the PR!

afrozenpeach:
Having the port in the host header is optional, it technically works, but can cause some problems in some setups. For my purposes, we use an HAProxy instance in front of the elasticsearch server. The HAProxy instance looks at the host header for routing, and having the port in the host header throws things off.

This is just a more proper way of doing things, and prevents some confusion and frustration when the host header is important.

Plus, having the port split out as it's own member variable is, in my opinion, cleaner than it is being in the host.

fullmoonissue:
Hello @polyfractal,

May I ping you about this PR ?

Is there anything missing before merging ?

Regards,

nicomro:
Hi @polyfractal ,
Any news ? I hope you are fine.
IMHO, it is still an issue.
Can this PR be merged ?
Regards,

nicomro:
Maybe someone else can handle this PR.
@ezimuel ?

ezimuel:
Thanks @nicomro for the remind. I'll review this PR asap and hopefully merge with 7.0.0 release.

ezimuel:
Just merged in master for 7.0.0 release. Thanks @afrozenpeach for this PR!

PR #792, opened 2018-08-25
When the custom handler, response['body'] is string, It throw warning.

limingxinleo:


ezimuel:
@limingxinleo can you provide a unit test for this issue? Thanks.

PR #800, opened 2018-09-15
fix phpdoc about versioning in update() and delete()

halaei:
Based on the final note in https://www.elastic.co/guide/en/elasticsearch/reference/2.2/docs-update.html, I am sending this to 5.0, which can be cerry picked to 6.0 and master as well.

Also 'version_type' can be removed from Update::getParamWhitelist() in master.


halaei:
I don't get any feedback.

PR #810, opened 2018-10-15
Adding 'pipeline' to UpdateByQuery.php whitelist

feydan:
Elasticsearch Update By Query supports the pipeline parameter.  This allows the parameter to be used.


autoferrit:
This would mean we don't need to use a couple workarounds that we do now in production.

polyfractal:
:+1: thanks!  I'm just about to run out the door, so I'll merge this and then tag a release first thing tomorrow morning when I'm back

florianajir:
When do you plan to tag a new release ?

PR #817, opened 2018-10-29
Fixed endpoint url for templates.

8lall0:

### DIFF
--- a/src/Elasticsearch/Endpoints/Template/Delete.php
+++ b/src/Elasticsearch/Endpoints/Template/Delete.php
@@ -28,7 +28,7 @@ class Delete extends AbstractEndpoint
             );
         }
         $templateId   = $this->id;
-        $uri  = "/_search/template/$templateId";
+        $uri  = "/_template/$templateId";
 
         return $uri;
     }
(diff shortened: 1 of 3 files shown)


feydan:
You may be looking for indices() template: https://github.com/elastic/elasticsearch-php/tree/5.0/src/Elasticsearch/Endpoints/Indices/Template, which is implemented correctly for index templates: https://www.elastic.co/guide/en/elasticsearch/reference/5.4/indices-templates.html.

It seems like the code you are changing here refers to search templates: https://www.elastic.co/guide/en/elasticsearch/reference/5.4/search-template.html

8lall0:
Yeah, you're right.
But i don't understand why where is a putTemplate() method inside the client, if it refers to search.

feydan:
Yea it's a little confusing - the search end points don't have their own namespace like indices, ingest, cluster, etc..., so the functions work directly off of the client.

Looks like the idea behind namespaces is that they provide more administrative functionality: https://www.elastic.co/guide/en/elasticsearch/client/php-api/5.0/_namespaces.html

PR #811, opened 2018-10-16
Upgrade phpstan to 0.11.4

akondas:
Next step is bump level to 3 :fearful: :rocket:

PR #821, opened 2018-11-19
Update to Mockery 1

carusogabriel:


carusogabriel:
Depends on #811, as `phpstan/phpstan-mockery` extension requires PHPStan 0.10

PR #822, opened 2018-12-02
remove default port 9200;

shaowangzhen:
don't want to set port when use domain;


polyfractal:
See comment in: https://github.com/elastic/elasticsearch-php/pull/309#issuecomment-149709983

Basically, the client is supposed to remain un-opinionated and reflect the defaults of Elasticsearch.  ES uses port 9200 (not port 80), so it should default to that.  If you want to use port 80, you'll need to set it explicitly, or write a wrapper around the client to do that by default for you.

PR #823, opened 2018-12-10
Remove FILTER_FLAG_SCHEME_REQUIRED for PHP 7.3

ian-nisbet:
PHP 7.3 generates a deprecation warning. Referring to https://wiki.php.net/rfc/deprecations_php_7_3#filter_flag_scheme_required_and_filter_flag_host_required

PR #803, opened 2018-09-25
Remove deprecated redundant FILTER_VALIDATE_URL from ClientBuilder::prependMissingScheme

alexcarol:
The intent of this PR is to enable compatibility with php 7.3.
Fixes https://github.com/elastic/elasticsearch-php/issues/798

PR #825, opened 2018-12-12
Fixed php 7.3 compatibiltiy for elasticsearch 2 / 1

### LINKED ISSUE #798: PHP 7.3
PHP 7.3 will deprecate FILTER_FLAG_SCHEME_REQUIRED and FILTER_FLAG_HOST_REQUIRED

https://wiki.php.net/rfc/deprecations_php_7_3#filter_flag_scheme_required_and_filter_flag_host_required

[src/Elasticsearch/ClientBuilder.php#L674
](https://github.com/elastic/elasticsearch-php/blob/7241e69095ce52c121fe2e9491fa2a7207be5dcf/src/Elasticsearch/ClientBuilder.php#L674)

```php
    /**
     * @param string $host
     *
     * @return string
     */
    private function prependMissingScheme($host)
    {
        if (!filter_var($host, FILTER_VALIDATE_URL, FILTER_FLAG_SCHEME_REQUIRED)) {
            $host = 'http://' . $host;
        }
        return $host;
    }
```

~~This function probably requires a php version check. In php 7.2 and later the flag can be omitted~~ It's 5.2, not 7.2. Just remove it

alexander-schranz:
Fixed php 7.3 compatibiltiy for elasticsearch 2 / 1 and fix travis ci problems with forked symfony yaml.

See #826 for Elasticsearch 5
See  #827   for Elasticsearch 6

See also #823 #803

https://wiki.php.net/rfc/deprecations_php_7_3#filter_flag_scheme_required_and_filter_flag_host_required

fixes #798

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ branches:
 sudo: false
 
 php:
+  - 7.3
+  - 7.2
+  - 7.1
   - 7.0
   - 5.6
   - 5.5
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ before_install:
   - ./travis/download_and_run_es.sh
 
 install:
-  - composer install --prefer-source
+  - composer install --prefer-dist
 
 before_script:
   - if [ $TRAVIS_PHP_VERSION = '7.0' ]; then PHPUNIT_FLAGS="--coverage-clover ./build/logs/clover.xml"; fi
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
   "repositories": [
     {
       "type": "vcs",
-      "url": "https://github.com/polyfractal/Yaml"
+      "url": "https://github.com/polyfractal/Yaml.git"
     }
   ]
 }
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
   "repositories": [
     {
       "type": "vcs",
-      "url": "https://github.com/polyfractal/Yaml.git"
+      "url": "https://github.com/polyfractal/Yaml"
     }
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -33,7 +36,7 @@ before_install:
   - ./travis/download_and_run_es.sh
 
 install:
-  - composer install --prefer-source
+  - composer install --prefer-dist
```
alexander-schranz: source make no sense in ci and will end up in errors like the previous commits: https://travis-ci.org/elastic/elasticsearch-php/jobs/467216401

polyfractal:
Hmm, I don't think we should probably merge this one.  ES 2.x is EOL and has been for a long time.  The client follows the same support cycle as core ES, so technically ES-PHP 2.x is EOL too.

alexander-schranz:
@polyfractal first thank you for merging the other pull requests!

For ES 2.0 you would do us a huge favor if this fix would also be merged and released. It would avoid implementing hacks in our CI Tests and allow to upgrade some servers PHP which still use an old ES cluster which currently cant be migrated to an newer versions.

polyfractal:
Ok, but just this once :)

I don't want to set a precedent (2.x really is EOL) but in this case it seems pretty innocuous to backport and release.  Y'all should really upgrade those servers though :P

Thanks for the PRs!

alexander-schranz:
@polyfractal Thank you!

polyfractal:
Np :)  Tidying up some other things in master/6.x, then will do a big set of tagging for all the branches

graste:
Thank you, @polyfractal

PR #861, opened 2019-04-02
Update README.md

lucasjkr:
Updated Documentation links to point to the correct URL

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Version Matrix
 
 Documentation
 --------------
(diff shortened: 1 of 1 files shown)


ezimuel:
Hi @lucasjkr, I'm the new maintainer for this library. I just close this in favor of #862. Thanks for your contribution!

lucasjkr:
Hi,

No worries, but #862 only fixes the bad link in one place, not both places. There's another link at the end of README that also points to a dead page that was missed in that one.

PR #862, opened 2019-04-02
Fix full docs link

hannesvdvreken:

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Version Matrix
 
 Documentation
 --------------
-[Full documentation can be found here.](http://www.elasticsearch.org/guide/en/elasticsearch/client/php-api/5.0/index.html)  Docs are stored within the repo under /docs/, so if you see a typo or problem, please submit a PR to fix it!
+[Full documentation can be found here.](https://www.elastic.co/guide/en/elasticsearch/client/php-api/6.0/index.html)  Docs are stored within the repo under /docs/, so if you see a typo or problem, please submit a PR to fix it!
 
 Installation via Composer
 -------------------------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,7 +43,7 @@ Version Matrix
 
 Documentation
 --------------
-[Full documentation can be found here.](http://www.elasticsearch.org/guide/en/elasticsearch/client/php-api/5.0/index.html)  Docs are stored within the repo under /docs/, so if you see a typo or problem, please submit a PR to fix it!
+[Full documentation can be found here.](https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/index.html)  Docs are stored within the repo under /docs/, so if you see a typo or problem, please submit a PR to fix it!
```
ezimuel: I think is better to use `6.0` instead of `current` in the URL, since here the README is related to version 6.x of Elasticsearch. Current is related to the latest version, that will become 7.0 soon.

ezimuel:
Hi @hannesvdvreken I'm the new maintainer of this library. I just provided feedback on your PR. Let me know what do you think, thanks!

README.md
hannesvdvreken: tbh i think current is better because it will never be outdated, unless you have a checklist ready to change this link to 7.0 before releasing the next version.

hannesvdvreken:
Hi @elizabethwann! I believe we attended some conferences together in the past. Good to hear you're taking the role of lead maintainer :-) wish you all the best.

README.md
ezimuel: I'm actually planning to release a new 6.x soon, that's why I'm asking.

hannesvdvreken:
@ezimuel fixed both links to have 6.0 references and squashed the commits. I think this is good to go now?

PR #870, opened 2019-04-18
Trim trailing slashes from hosts when passed as a string.

### LINKED ISSUE #869: Trailing slashes breaking some operations when using a credential provider
The exception is thrown in another library - https://github.com/jeskew/amazon-es-php - but maybe the fix should be made in this repository. Therefore I am raising an issue in this repository as well.

For reference, the issue in https://github.com/jeskew/amazon-es-php: https://github.com/jeskew/amazon-es-php/issues/17

When switching Elasticsearch cluster I copied the endpoint uri and received the uri including trailing slash. This led to failed calls for some operations due to an invalid uri exception.

### Reproduction scenario
I have been able to reproduce this issue using https://github.com/jeskew/amazon-es-php, but any handler that "simply" copies a path behind the host it receives might be affected.

```php
$handler = new \Aws\ElasticsearchService\ElasticsearchPhpHandler(
    'eu-west-1', 
    CredentialProvider::defaultProvider()
);
$client = \Elasticsearch\ClientBuilder::fromConfig([
    'hosts' => ['http://host.docker.internal:9200/'], 
    'handler' => $handler
], true);

$client->info()
```

This will lead to an `InvalidArgumentException` with the message "'Unable to parse URI: //'" to be triggered in the creation of a PSR-7 request at https://github.com/jeskew/amazon-es-php/blob/b2b5634e4662f5d620891bda9b6d0af73954043c/src/ElasticsearchPhpHandler.php#L63

The path of the uri is `/`, to get the information of the host `/` is added and thus the uri becomes `//`, which in itself is an invalid uri. I have only seem this issue occur with this operation, since the other operations add something to the path resulting in uris like `//my_index`. Those uris *are* excepted by the `Uri` constructor. 

The easiest solution might be to add a sanitization step to `\Elasticsearch\ClientBuilder::setHosts()` that removes trailing slashes. I am willing to look into making a PR to accomplish that.

### System details
- PHP 7.2
- `jsq/amazon-es-php` version `0.2.3` (latest stable)
- `elasticsearch/elasticsearch` version `6.1.0` (latest stable)
- OS: standard `php:7.2-fpm` Docker container

timoschinkel:
When a host contains trailing slashes this might lead to issues when requesting the path `/`,
as this might lead to the uri `//`. This is an invalid uri - `parse_url()` will return `false`
on this uri.

This PR only removes trailing slashes from hosts that are passed as a string. Rationale behind
this decision is that, when passed as an array, the path is a separate entry in that array.

Maybe this should be handled by the code that calls the `ClientBuilder` or maybe the handler
where this issue presented itself in this scenario, but I think by sanitizing the uri by
proactively removing the trailing slashes this library will become more robust.

This PR fixes #869

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -610,6 +610,7 @@ class ClientBuilder
         foreach ($hosts as $host) {
             if (is_string($host)) {
                 $host = $this->prependMissingScheme($host);
+                $host = $this->removeTrailingSlashes($host);
                 $host = $this->extractURIParts($host);
             } elseif (is_array($host)) {
                 $host = $this->normalizeExtendedHost($host);
@@ -676,5 +677,15 @@ class ClientBuilder
         }
 
         return $host;
+    }
+
+    /**
+     * @param string $host
+     *
+     * @return string
+     */
+    private function removeTrailingSlashes($host)
+    {
+        return rtrim($host, '/');
     }
 }


ezimuel:
@timoschinkel thanks for this PR but as I commented in #869 this is not an issue for elasticsearch-php library. I'm sorry but I cannot merge it. Again, thanks for your effort.

PR #877, opened 2019-04-23
Patch 4

armgabrielyan:


ezimuel:
@armgabrielyan what's the goal of this PR? I think you worked against an old `master` version. I'm closing it, until you will fix the issue and explain the goal. Thanks!

PR #884, opened 2019-04-26
Fix/6.7

ezimuel:
This PR contains the fix for Elasticsearch 6.7. I will release elasticsearch-php `6.7.0` after merging this PR.

### DIFF
--- a/src/Elasticsearch/Endpoints/Indices/Mapping/Put.php
+++ b/src/Elasticsearch/Endpoints/Indices/Mapping/Put.php
@@ -41,19 +41,21 @@ class Put extends AbstractEndpoint
      */
     public function getURI()
     {
-        if (isset($this->type) !== true) {
+        $index = $this->index ?? null;
+        $type = $this->type ?? null;
+
+        if (null === $index && $type === $index) {
             throw new Exceptions\RuntimeException(
-                'type is required for Put'
+                'type or index are required for Put'
             );
         }
-        $index = $this->index;
-        $type = $this->type;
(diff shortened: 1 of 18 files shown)


src/Elasticsearch/Endpoints/Indices/Mapping/Put.php
```diff
@@ -41,19 +41,21 @@ public function setBody($body)
...
     public function getURI()
     {
-        if (isset($this->type) !== true) {
+        $index = $this->index ?? null;
+        $type = $this->type ?? null;
+
+        if (null === $index && $type === $index) {
             throw new Exceptions\RuntimeException(
-                'type is required for Put'
+                'type or index are required for Put'
```
philkra: shouldn't it be `'type and index are required for Put'`?

philkra:
LGTM, besides the possible typo, of which I'm not sure of :)

src/Elasticsearch/Endpoints/Indices/Mapping/Put.php
ezimuel: Actually, in ES 6.7 the {type} part is not required anymore. If you see the available paths [here](https://github.com/elastic/elasticsearch/blob/v6.7.0/rest-api-spec/src/main/resources/rest-api-spec/api/indices.put_mapping.json#L6-L7) is it possible to request `{index}/_mapping`, `/_mapping/{type}`, `{index}/_mappings/{type}`, etc.

PR #885, opened 2019-04-29
Index closed exception

gempir:
Added an IndexClosedException to catch this error more easily.

For reference this is a index_closed_exception response from elasticsearch.

```
{
  "error": {
    "root_cause": [
      {
        "type": "index_closed_exception",
        "reason": "closed",
        "index_uuid": "bNLHYp4UQNu3ZF-YSJ1awg",
        "index": "navigation_index_2"
      }
    ],
    "type": "index_closed_exception",
    "reason": "closed",
    "index_uuid": "bNLHYp4UQNu3ZF-YSJ1awg",
    "index": "navigation_index_2"
  },
  "status": 400
}
```


elasticcla:
Hi @gempir, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-php/pull/885.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Issue #831, opened 2019-01-13
Error: Cannot use object of type Elasticsearch\Common\Exceptions\Missing404Exception as array


yochanan-gurovich:
Trying to get a document by an ID that does not exist.
Code is in a try / catch block, and is caught.
The above error is thrown still.

Code snippet (username, password and host are masked):
```php
        try {
            $hosts = [
                'https://username:password@host:port'
            ];
            $client = Elasticsearch\ClientBuilder::create()->setHosts($hosts)->build();
            $params = [
                'index' => 'products',
                'type' => 'default',
                'id' => 'non_existant_id'
            ];
            $response = $client->get($params);
            var_dump($response);
        } catch (\Exception $e) {
            echo $e->getMessage();
        }
```

- Operating System: Ubuntu 18.04
- PHP Version: 7.2
- ES-PHP client version: 6.1.0
- Elasticsearch version: 6.5.4 (Elastic Cloud)

polyfractal:
Can you paste the full exception stack, so I can see where it's originating from?

yochanan-gurovich:
#0 /home/forge/6day.store/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Connections/Connection.php(279): Elasticsearch\Connections\Connection->process4xxError(Array, Array, Array)
#1 /home/forge/6day.store/vendor/react/promise/src/FulfilledPromise.php(25): Elasticsearch\Connections\Connection->Elasticsearch\Connections\{closure}(Array)
#2 /home/forge/6day.store/vendor/guzzlehttp/ringphp/src/Future/CompletedFutureValue.php(55): React\Promise\FulfilledPromise->then(Object(Closure), NULL, NULL)
#3 /home/forge/6day.store/vendor/guzzlehttp/ringphp/src/Core.php(341): GuzzleHttp\Ring\Future\CompletedFutureValue->then(Object(Closure), NULL, NULL)
#4 /home/forge/6day.store/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Connections/Connection.php(299): GuzzleHttp\Ring\Core::proxy(Object(GuzzleHttp\Ring\Future\CompletedFutureArray), Object(Closure))
[...]

PR #889, opened 2019-05-06
Fix/rejection handler failed

antismok:
[Error: Cannot use object of type Elasticsearch\Common\Exceptions\Missing404Exception as array](https://github.com/elastic/elasticsearch-php/issues/831)


elasticcla:
Hi @antismok, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-php/pull/889.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #890, opened 2019-05-06
Fix/rejection handler failed

antismok:
FIX [Error: Cannot use object of type Elasticsearch\Common\Exceptions\Missing404Exception as array](https://github.com/elastic/elasticsearch-php/issues/831)

### DIFF
--- a/src/Elasticsearch/Transport.php
+++ b/src/Elasticsearch/Transport.php
@@ -194,7 +194,7 @@ class Transport
      *
      * @return bool
      */
-    protected function ignoreTheError(?int $errorCode): bool
+    protected function ignoreTheError($errorCode = null): bool
     {
         if (null === $errorCode) {
             return false;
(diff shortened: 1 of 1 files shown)


elasticcla:
Hi @antismok, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-php/pull/890.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

antismok:
> @antismok sorry if I come late to this PR. I closed the issue #831 because I cannot reproduce the error. I used the code example reported in my comment [#831 (comment)](https://github.com/elastic/elasticsearch-php/issues/831#issuecomment-520778823).
> Is this still an issue for you? Can you elaborate more? Thanks.

Еhere are no problems with ELK version 6, problems with version ELK 7+

ezimuel:
I'm closing this issue since I cannot reproduce it, see https://github.com/elastic/elasticsearch-php/issues/831#issuecomment-520778823

Issue #891, opened 2019-05-10
Tests broken on ElasticSearch php v6.7.0


anam-hossain:
### Summary of problem or feature request

Unit tests broken on ElasticSearch PHP v6.7
```
 "name": "elasticsearch/elasticsearch",
 "version": "v6.7.0",
```

Mock handler not working anymore.

### Code snippet of problem
```php
use GuzzleHttp\Ring\Client\MockHandler;
use Elasticsearch\ClientBuilder;

$response = json_encode([
            "_id" => 100,
            "result" => "created",
        ]);

$handler = new MockHandler([
            'status' => $status,
            'transfer_stats' => [
                'total_time' => 100,
            ],
            'body' => fopen('data://text/plain,' . $response, 'r'),
        ]);

$builder = ClientBuilder::create();
$builder->setHosts(['localhost']);
$builder->setHandler($handler);

$client = $builder->build();

// Indexing
 $params = [
            'index' => 'foo',
            'type' => 'bar',
            'id' => 100,
            'body' => ['name' => 'James'],
        ];

$client->index($params);
```
When run it, it returns

```
TypeError: strpos() expects parameter 1 to be string, null given
```
After debugging I found that at line 568 in Connection.php, the $uri returning null where its expecting string.

```
 private function buildCurlCommand($method, $uri, $body)
    {
     // Uri null  
      if (strpos($uri, '?') === false) {
            $uri .= '?pretty=true';
        } else {
            str_replace('?', '?pretty=true', $uri);
        }

        $curlCommand = 'curl -X' . strtoupper($method);
        $curlCommand .= " '" . $uri . "'";

        if (isset($body) === true && $body !== '') {
            $curlCommand .= " -d '" . $body . "'";
        }

        return $curlCommand;
    }
```
### System details

- Mac OS 10.14.4
- PHP 7.3
- ES-PHP client version: v6.7.0

CheerfulPlum:
Hey @anam-hossain. Have found a workaround for this. Passing in a effective_url key into the MockHandler constructor will fix the issue and it'll return your body when you call index (or in my case search).
```php
$handler = new MockHandler([
'status' => $status,
            'transfer_stats' => [
                'total_time' => 100,
            ],
            'body' => fopen('data://text/plain,' . $response, 'r'),
            'effective_url' => 'localhost'
        ]);
```

@ezimuel should I do a pull request to update the README?

anam-hossain:
@CheerfulPlum Thank you very much.

[...]

PR #895, opened 2019-06-07
#891 Update README for effective_url with MockHandler

CheerfulPlum:
Update documentation to fix the problem detailed in #891


elasticcla:
Hi @CheerfulPlum, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-php/pull/895.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #896, opened 2019-06-07
Update README for effective_url with MockHandler #895

CheerfulPlum:
Update documentation to fix the problem detailed in #891


ezimuel:
Thanks @CheerfulPlum

PR #899, opened 2019-06-14
Wait for the future response

marcvdm:
The current `SniffingConnectionPool` uses the sniff method on the `Connection` class but this needs to return an array. The current implementation fails because a future request is returned.

This merge requests adds the wait function for the return.

### DIFF
--- a/src/Elasticsearch/ConnectionPool/SniffingConnectionPool.php
+++ b/src/Elasticsearch/ConnectionPool/SniffingConnectionPool.php
@@ -124,7 +124,7 @@ class SniffingConnectionPool extends AbstractConnectionPool implements Connectio
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


ezimuel:
The `Connection::performRequest()` returns a mixed type. When I did the code refactoring, introducing type hints and return types, I struggled between the old implementation and the new one.
Maybe a better idea is to remove the return type on the `sniff()` function, in order to preserve asynchronous feature. What do you think? Thanks for reporting this issue.

marcvdm:
I'v updated the pull request and remove the typehints. This way the response can be an array or FutureArray and the parseClusterState wil iterate over an `array` or `ArrayAccess` variable

ezimuel:
Thanks @marcvdm

PR #926, opened 2019-08-06
Doc Examples

philkra:
* final directory structure
* first stab at generator


examples/docs/generator/generate.php
```diff
@@ -0,0 +1,199 @@
...
+    $asciidir = sprintf('../asciidoc/%s%s/', $folders, $filename);
+
+    // Build Sub directory Structure
+    if(file_exists($asciidir) === false) {
+        mkdir($asciidir, 0775, true);
+    }
+
+    // Generate Asciifdoc Include Files
+    foreach($examples as $example) {
+        if(file_exists(sprintf('%s/%s.asciidoc', $asciidir, $example['hash'])) === false && file_exists(sprintf('%s/_%s.asciidoc', $asciidir, $example['hash'])) === false) {
```
nik9000: So you make a `_` file which will resolve the include and just remove the `_` when you are ready. Clever.

nik9000: I'm starting to prefer the `.adoc` suffix over `.asciidoc`. Both are standard to be honest and you can totally do whichever.

examples/docs/generator/helpers.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+// === getting-started.asciidoc: line 209: f8cc4b331a19ff4df8e4a490f906ee69
```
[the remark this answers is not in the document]
philkra: yeah, the use of the whole file is yet to be determined or to say it frankly, I will delete it :)

examples/docs/src/Examples/Docs/Aggregations/Bucket/AdjacencyMatrixAggregation.php
```diff
@@ -12,7 +12,7 @@
  *
  * Class: AdjacencyMatrixAggregation
  *
- * Date: 2019-08-05 08:49:19
+ * Date: 2019-08-06 06:56:02
```
ezimuel: Is the date important here? If we run the generator and the code is the same we'll have different files with just a date diff.  This is something expected or can we remove it to avoid superfluous commits?

examples/docs/generator/generate.php
philkra: your choice, will rename the files to `.adoc`

examples/docs/src/Examples/Docs/Aggregations/Bucket/AdjacencyMatrixAggregation.php
nik9000: I do like the idea of keeping the diff small!

philkra: no it's not, you are 100% correct on the diff, that bit me already a couple of times. will remove it

philkra:
Summary:
* add new line to asciidoc skeleton
* changed the file extension to `.adoc` in the generator and the existing files
* removed the timestamp from the classes and the methods
* deleted the incredibly useless `helpers.php` file :)

PR #938, opened 2019-08-21
Missing Implementation for Doc Examples

philkra:
The Get API page had a missing implementation for the put mapping code snippet.


philkra:
Yes, it _bothers_ me as well! Let me rework this to append a `Page` to every class name, e.g. `GetPage`. Or do you have a better suggestion for semantically reasonable naming suffix?

ezimuel:
Proposal: use the same naming convention that we have for the endpoints. We will have small classes (unit tests) for each endpoint and this will also facilitate the search for examples. @philkra what do you think?

ezimuel:
We discussed in a meeting with @philkra and it's ok for me to follow the actual naming convention for the doc examples. LGTM!

PR #949, opened 2019-09-18
Fix prependMissingScheme with underscore

ihor-sviziev:
While working with ElasticSearch in docker using docker-compose - hostname for my ES was `elasticsearch.myproject_default`. I tried to use it with this ES client - it failed to connect. 

I configured hostname to `http://elasticsearch.myproject_default:9200`. After changing to `elasticsearch.myproject_default:9200` my issue was fixed, but that wasn't expected at all

Deeper investigation showed that issue caused by incorrect URL validation. More info: https://stackoverflow.com/questions/39539468/php-filter-validate-url-not-finding-subdomains-with-underscore

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -686,7 +686,7 @@ class ClientBuilder
 
     private function prependMissingScheme(string $host): string
     {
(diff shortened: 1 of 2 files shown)


ezimuel:
Thanks @ihor-sviziev, good catch!

PR #952, opened 2019-09-26
Mutator for X-Opaque-Id Header

philkra:
This PR adds support for simplified access to the `X-Opaque-Id` header.


ezimuel:
Thanks @philkra!

PR #957, opened 2019-10-10
[types] add @var annotatoins

TomasVotruba:
This might help IDE for better usability of code


ezimuel:
@TomasVotruba thanks for your contribution!

PR #964, opened 2019-11-15
Automated native_function_invocation fixes

draco2003:
Reproduce with:
php php-cs-fixer --rules=native_function_invocation fix src --allow-risky=yes
(php-cs-fixer from : https://cs.symfony.com/)

Details:
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

Will get CLA setup next week if we don't already have one.


elasticcla:
Hi @draco2003, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-php/pull/964.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

ezimuel:
@draco2003 this is something I noted in my TODO list months ago and planned for `8.0` milestone. Instead of using global namespace for native function (e.g. `\array_key_exists()`), as you proposed, I would like to use namespace. For instance:
```php
namespace Elasticsearch;

use Elasticsearch\Namespaces\CatNamespace;
// ...
use function array_key_exists;

class Client
{
    // ...
    public function extractArgument(array &$params, string $arg)
    {
        if (array_key_exists($arg, $params) === true) {
            // ...
        }
    }
}
```

Can you refactor the PR to use namespaces? Thanks!

ezimuel:
I just opened #965 as remainder for `8.0` milestone.

PR #968, opened 2019-11-20
Fix pre 7.2 endpoint class name with aliases + reapply fix #947

### LINKED ISSUE #967: Compatibility issues with version 7.4.0

Since version 7.4.0 was published, it seems some classes were renamed which break compatiblity.

For example, `Elasticsearch\Endpoints\Ingest\Pipeline\Delete` was renamed to `Elasticsearch\Endpoints\Ingest\DeletePipeline`

See https://github.com/ruflin/Elastica/issues/1710 for reference.

ezimuel:
This PR fixes #967 introducing a class alias to all the previous endpoint namespaces present in `elasticsearch-php 7.2`.

I generated the class name for all the endpoints in 7.2 branch and I produced an alias in `src/autoload.php` (I removed the outdated endpoints, not available anymore in Elasticsearch 7).
I added the autoload.php file in composer.
**NOTE:** This file will be removed in the next major version (8.0.0).

I also reapplied the fix #947 for `IndicesNamespace::getAliases()`.


src/autoload.php
```diff
@@ -0,0 +1,49 @@
...
+class_alias('\Elasticsearch\Endpoints\PutScript', '\Elasticsearch\Endpoints\Script\Put');
+class_alias('\Elasticsearch\Endpoints\DeleteScript', '\Elasticsearch\Endpoints\Script\Delete');
+class_alias('\Elasticsearch\Endpoints\Snapshot\CreateRepository', '\Elasticsearch\Endpoints\Snapshot\Repository\Create');
+class_alias('\Elasticsearch\Endpoints\Snapshot\DeleteRepository', '\Elasticsearch\Endpoints\Snapshot\Repository\Delete');
+class_alias('\Elasticsearch\Endpoints\Snapshot\GetRepository', '\Elasticsearch\Endpoints\Snapshot\Repository\Get');
+class_alias('\Elasticsearch\Endpoints\Snapshot\VerifyRepository', '\Elasticsearch\Endpoints\Snapshot\Repository\Verify');
+class_alias('\Elasticsearch\Endpoints\GetSource', '\Elasticsearch\Endpoints\Source\Get');
+class_alias('\Elasticsearch\Endpoints\ExistsSource', '\Elasticsearch\Endpoints\Source\Exists');
+class_alias('\Elasticsearch\Endpoints\Tasks\ListTasks', '\Elasticsearch\Endpoints\Tasks\TasksList');
+class_alias('\Elasticsearch\Endpoints\Termvectors', '\Elasticsearch\Endpoints\TermVectors');
```
thePanz: The alias is not needed, same name on both sides

deguif: Case has changed. But I'm not sure if class names are case sensitive.

ezimuel: @thePanz is right here. The alias cannot be done for case sensitive, since classes are case insensitive in PHP (even if the autoloading of composer is case sensitive). I'll just reuse the old class naming convention here.

src/autoload.php
```diff
@@ -0,0 +1,47 @@
...
+ * File required since elasticsearch-php 7.4.1 for aliasing
+ * to the previous endpoint class names.
+ *
+ * @see https://github.com/elastic/elasticsearch-php/issues/967
+ *
+ * NOTE: This file will be removed with elasticsearch-php 8.0.0
+ */
+
+class_alias('\Elasticsearch\Endpoints\Nodes\Info', '\Elasticsearch\Endpoints\Cluster\Nodes\Info');
+class_alias('\Elasticsearch\Endpoints\Nodes\Stats', '\Elasticsearch\Endpoints\Cluster\Nodes\Stats');
```
deguif: What about previous classes:
- `Elasticsearch\Endpoints\Cluster\Nodes\HotThreads`
- `Elasticsearch\Endpoints\Cluster\Nodes\Usage`
- `Elasticsearch\Endpoints\Cluster\Nodes\ReloadSecureSettings`

ezimuel: You right, just added it. Thanks for this code review!

PR #972, opened 2019-11-25
!! fix $scroll_id condition

AymanElarian:
!! fix $scroll_id condition

### DIFF
--- a/src/Elasticsearch/Endpoints/Scroll.php
+++ b/src/Elasticsearch/Endpoints/Scroll.php
@@ -23,7 +23,7 @@ class Scroll extends AbstractEndpoint
     public function getURI(): string
     {
         $scroll_id = $this->scroll_id ?? null;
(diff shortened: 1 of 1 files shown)


ezimuel:
@AymanElarian actually the original code is ok. The `E_USER_DEPRECATED` notice is sent when you set the `scroll_id`, according to the Elasticsearch API specification (see [here](https://github.com/elastic/elasticsearch/blob/v7.4.2/rest-api-spec/src/main/resources/rest-api-spec/api/scroll.json#L24-L32)).

Basically, the `scroll_id` should be set in the body of the request, as follows:

```php
$client = ClientBuilder::create()->build();
$client->scroll(
    'body' => [
        'scroll_id' => /* ... */
    ]
);
```

AymanElarian:
i see . , thank you for clarification

PR #983, opened 2019-12-17
Fix extractArgument iterable casting to array

rtyshyk:
Bulk API suports data as `itarable`, see ` \Elasticsearch\Endpoints\Bulk::setBody`.

It has been broken during refactoring, see  #56163ff.

### DIFF
--- a/src/Elasticsearch/Client.php
+++ b/src/Elasticsearch/Client.php
@@ -1275,7 +1275,9 @@ class Client
     {
         if (array_key_exists($arg, $params) === true) {
             $value = $params[$arg];
-            $value = is_object($value) ? (array) $value : $value;
+            $value = (is_object($value) && !is_iterable($value)) ?
+                (array) $value :
+                $value;
             unset($params[$arg]);
             return $value;
         } else {
--- a/tests/Elasticsearch/Tests/ClientTest.php
+++ b/tests/Elasticsearch/Tests/ClientTest.php
@@ -478,4 +478,16 @@ class ClientTest extends \PHPUnit\Framework\TestCase
         $result = $client->info($params);
         $this->assertInstanceOf(FutureArray::class, $result);
     }
+
+    public function testExtractArgumentIterable()
+    {
+        $client = Elasticsearch\ClientBuilder::create()->build();
+        // array iterator can be casted to array back, so make more real with IteratorIterator
+        $body = new \IteratorIterator(new \ArrayIterator([1, 2, 3]));
+        $params = ['body' => $body];
+        $argument = $client->extractArgument($params, 'body');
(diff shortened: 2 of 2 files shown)


rtyshyk:
I signed, but not sure if / how to update CLA status in checks.

elasticcla:
Hi @rtyshyk, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-php/pull/983.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

rtyshyk:
> Hi @rtyshyk, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/elasticsearch-php/pull/983.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

added second email as well.

ezimuel:
@rtyshyk what is the issue that you are trying to solve? The Bulk endpoint supports [Traversable](https://www.php.net/manual/en/class.traversable.php) (and also [Iterator](https://www.php.net/manual/en/class.iterator.php) that extends Traversable) using [this code](https://github.com/elastic/elasticsearch-php/blob/v7.4.1/src/Elasticsearch/Endpoints/Bulk.php#L71-L74).
The fact that `Client::extractArgument()` converts always an object into an array I don't think is an issue for the endpoint, even because this function is used also by other endpoints and not just `Bulk`. The `Traversable` support in `Bulk` endpoint has been added in https://github.com/elastic/elasticsearch-php/pull/351.

ezimuel:
> I signed, but not sure if / how to update CLA status in checks.

The CLA signature is ok now.

rtyshyk:
@ezimuel Yes, bulk endpoint supports `Traversable`, but the issue here is it is not possible to pass `Traversable` to it :)

I use `\Elasticsearch\Client` to consume `bulk` endpoint and pass there Iterator that returns right meta/data to build a body as well as reduce memory usage. 

to make it simple
```php
$client = new \Elasticsearch\Client()
$result = $client->bulk([
    'body' => new \IteratorIterator(new \ArrayIterator([1, 2, 3])),
]);
```

`\Elasticsearch\Client::bulk` uses `extractArgument` for the `body` parameter where is  extractArgument do a cast any object to array.

Traversable cannot be cast to an array, as a result, there is an empty data argument.

ezimuel:
@rtyshyk I see what you mean, I wrongly supposed that casting `IteratorIterator` into `array` was working fine. I'll merge this PR soon.
Thanks for your contribution!

PR #997, opened 2020-02-14
Fix for #993, added ClientBuilder::includePortInHostHeader()

### LINKED ISSUE #993: Port is not properly set in requests, causing No alive nodes found in your cluster

The elasticsearch-php library does not set the correct Host header when using a custom port.

I was getting "No alive nodes found in your cluster" when performing a query using elasticsearch-php, but everything was working fine when doing the same request using curl on the command line.

After setting `CURLOPT_VERBOSE => 1` in the script and using the `-v` flag in the CLI, I noticed that the HTTP `Host` header  was what differed between the elasticserach-php request and the CLI request. The CLI request set the Host header to `elastic.example.com:9243` but the elasticsearch-php client set the Host header to `elastic.example.com`, even though the request was still made to the correct port on the TCP layer. 

The missing port number in the Host header led to the server sending an empty response (cURL error 52: Empty reply from server).

The elasticsearch cluster is running HTTPS, on port 9243, with HTTP basic auth. This bug applies both when configuring the host (`ClientBuilder::setHosts()`) using a string such as `https://user:pass@elastic.example.com:9243` as well as when configuring the host with an array containing the same information in separate fields (scheme, host, port, etc).

**Edit:** After looking a bit closer at the other tickets (and having lunch...), I realized that this is closely related to #925, which also explains that this bug was introduced in #782. The author of #782 says that "having the port in the host header is optional", but as @polyfractal says a bit earlier in the same thread [this is not really the case](https://github.com/elastic/elasticsearch-php/issues/548#issuecomment-282495025) if you read [the RFC](https://tools.ietf.org/html/rfc2616#page-128). Elasticsearch uses the HTTP protocol which means the implied port will be 80, or in the case of HTTPS 443. You cannot omit it and still stay compliant with HTTP/S, and if you're not compliant with HTTP/S then all bets are off when it comes to configuration of proxies, web servers, etc, which invalidates the reason for introducing the PR in the first place.

### Code snippet of problem

The problem is fixed by changing line 208 in `Elasticsearch\Connections\Connection` (in the `performRequest` method) from

    'Host' => [$this->host]

to

    'Host' => ["{$this->host}:{$this->port}"]

### System details

- Operating System: Ubuntu 16.04
- PHP Version: 7.4.2
- ES-PHP client version: 7.4.1
- Elasticsearch version: 7.5.0

ezimuel:
This PR fixes #993 adding a `ClientBuilder::includePortInHostHeader(bool $enable)`. Using this option the HTTP port will be added to `Host` header if it's not standard (80 for HTTP, 443 for HTTPS).

For instance, this configuration will build the header `Host: localhost:9200`.
```php
$host = 'localhost:9200';
    ->setHosts([$host])
    ->includePortInHostHeader(true)
    ->build();
```
If the port is standard (e.g. 80) even if `includePortInHostHeader(true)` will not include the port in the `Host` header.

By default (if you don't call `includePortInHostHeader(true)`), the port is not included for backward compatibility.

### DIFF
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -200,9 +200,7 @@ class Connection implements ConnectionInterface
 
         $host = $this->host;
         if (isset($this->connectionParams['client']['port_in_header']) && $this->connectionParams['client']['port_in_header']) {
-            if (!in_array((int) $this->port, [80,443])) {
-                $host .= ':' . $this->port;
-            }
+            $host .= ':' . $this->port;
         }
         
         $request = [
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -168,6 +168,11 @@ class Connection implements ConnectionInterface
         }
         $port = $hostDetails['port'];
 
+        if (isset($connectionParams['client']['port_in_header']) && $connectionParams['client']['port_in_header']) {
+            if (!in_array((int) $port, [80,443])) {
+                $host .= ":$port";
+            }
+        }
         $this->host             = $host;
         $this->path             = $path;
         $this->port             = $port;
@@ -198,6 +203,7 @@ class Connection implements ConnectionInterface
             $this->headers = array_merge($this->headers, $options['client']['headers']);
         }
 
+        
         $request = [
             'http_method' => $method,
             'scheme'      => $this->transportSchema,
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -168,11 +168,6 @@ class Connection implements ConnectionInterface
         }
         $port = $hostDetails['port'];
 
-        if (isset($connectionParams['client']['port_in_header']) && $connectionParams['client']['port_in_header']) {
-            if (!in_array((int) $port, [80,443])) {
-                $host .= ":$port";
-            }
-        }
         $this->host             = $host;
         $this->path             = $path;
         $this->port             = $port;
@@ -203,6 +198,12 @@ class Connection implements ConnectionInterface
             $this->headers = array_merge($this->headers, $options['client']['headers']);
         }
 
+        $host = $this->host;
+        if (isset($this->connectionParams['client']['port_in_header']) && $this->connectionParams['client']['port_in_header']) {
+            if (!in_array((int) $this->port, [80,443])) {
+                $host .= ':' . $this->port;
+            }
+        }
         
         $request = [
             'http_method' => $method,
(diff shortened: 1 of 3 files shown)


ezimuel:
@larsnystrom may you review this PR? Thanks!

src/Elasticsearch/Connections/Connection.php
```diff
@@ -168,6 +168,11 @@ public function __construct(
         }
         $port = $hostDetails['port'];
 
+        if (isset($connectionParams['client']['port_in_header']) && $connectionParams['client']['port_in_header']) {
+            if (!in_array((int) $port, [80,443])) {
```
larsnystrom: I think it should be possible to have port 80 or 443 in the header as well. It is perfectly acceptable according to spec, so I don't see a need for special treatment.

src/Elasticsearch/Connections/Connection.php
```diff
@@ -168,6 +168,11 @@ public function __construct(
...
+                $host .= ":$port";
```
larsnystrom: Instead of appending the port here in the constructor, I think you should append when a request is made in `performRequest()`. The setting is called `include_port_in_host_header`, so it makes sense to only include it in the header, and the header is only constructed in `performRequest`

larsnystrom: (If you do it in `performRequest()` instead it is possible to get just the host part when you call `getHost()`. Those that need to get the port as well have a `getPort()` method for that. But the way its implemented right now makes it impossible to get the original `host` part back from the `Connection` object.)

src/Elasticsearch/Connections/Connection.php
larsnystrom: Including the port is ok according to spec, and those who do not want the port in the Host header can just set the newly introduced option to false to disable it.

ezimuel: I removed the default ports because it's a common standard in many HTTP client. Moreover, I suspect this will fix also #782.

src/Elasticsearch/Connections/Connection.php
ezimuel: Very good point! I'm refactoring it as you suggested. Thanks!

src/Elasticsearch/Connections/Connection.php
larsnystrom: I don't feel super strongly about this, but my reasoning goes like this:
1. The elasticsearch client adds the port to the Host header, as is allowed by the HTTP spec.
2.  Somebody will encounter #548, for which the easy solution is to not append the port to the Host header (they could also fix the broken proxy server, but lets be realistic).
3.  They can accomplish this with `includePortInHostHeader(false)`.

So this PR solves #782 even if you don't exclude port 80 and 443.

Anyway, you decide :)

ezimuel: :thinking: Ok, you convinced me. I'll remove this standard port check from the PR. I'll investigate more on that for the next major version of elasticsearch-php. Thanks again for your precious feedback.

PR #1016, opened 2020-04-09
Better exception message for Could not parse URI

alexander-schranz:
Not sure what the current lowest supported version is. Please change the target branch or ping me.

Most users run into this exception when they configure the URI false but mostly we are not sure if the ENV environment don't work or if they just really configured a not parse able url. So it would be great to output the given value in the exception message also to improve developer experience.

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -670,7 +670,7 @@ class ClientBuilder
         $parts = parse_url($host);
 
         if ($parts === false) {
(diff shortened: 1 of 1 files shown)


ezimuel:
@alexander-schranz thanks for this PR. I think I need to merge this also in `6.x`, `7.x` and `master`. I'll start releasing next week with `7.7`.

ezimuel:
I merged in 5.x, 6.x, 7.x and master. Ready for 7.7 to be released.

PR #1036, opened 2020-05-25
Make ClientBuilder override possible

snoob:
As instantiate method is protected, using static instead of self allow you to create your own ClientBuilder and override it


ezimuel:
@snoob thanks for this PR. I think using `static` will simplify the overriding of ClientBuilder.

ezimuel:
Just released `elasticsearch-php` [7.9.0](https://github.com/elastic/elasticsearch-php/releases/tag/v7.9.0) that includes #1036.

PR #1051, opened 2020-07-08
Fixes #1042 Fix PHP preloading

### LINKED ISSUE #1042: Elasticsearch-php produces lots of warnings with php 7.4 preloading enabled.

I use elasticsearch-php within a symfony project (symfony 4.4). Elasticsearch-php produces lots of warnings with php 7.4 preloading enabled.

Example log line:
NOTICE: PHP message: PHP Warning:  Cannot declare class \Elasticsearch\Endpoints\Cluster\Nodes\HotThreads, because the name is already in use in /code/xx/vendor/elasticsearch/elasticsearch/src/autoload.php on line 11

This line is repeated for every class_alias defined in the mentioned autoload.php

### System details

- Operating System: Alpine Linux
- PHP Version: 7.4.6
- ES-PHP client version: 7.7.0
- Elasticsearch version: 7.7

annuh:
This PR should fix the following warning when using PHP autoloading with Symfony:

```
Warning: Cannot declare class \Elasticsearch\Endpoints\Cluster\Nodes\HotThreads, because the name is already in use in /var/www/html/vendor/elasticsearch/elasticsearch/src/autoload.php on line 11

Warning: Cannot declare class \Elasticsearch\Endpoints\Cluster\Nodes\Info, because the name is already in use in /var/www/html/vendor/elasticsearch/elasticsearch/src/autoload.php on line 12

Warning: Cannot declare class \Elasticsearch\Endpoints\Cluster\Nodes\ReloadSecureSettings, because the name is already in use in /var/www/html/vendor/elasticsearch/elasticsearch/src/autoload.php on line 13

Warning: Cannot declare class \Elasticsearch\Endpoints\Cluster\Nodes\Stats, because the name is already in use in /var/www/html/vendor/elasticsearch/elasticsearch/src/autoload.php on line 14

Warning: Cannot declare class \Elasticsearch\Endpoints\Cluster\Nodes\Usage, because the name is already in use in /var/www/html/vendor/elasticsearch/elasticsearch/src/autoload.php on line 15

Warning: Cannot declare class \Elasticsearch\Endpoints\Cluster\Settings\Get, because the name is already in use in /var/www/html/vendor/elasticsearch/elasticsearch/src/autoload.php on line 16
```

### DIFF
--- a/src/autoload.php
+++ b/src/autoload.php
@@ -8,43 +8,51 @@
  * NOTE: This file will be removed with elasticsearch-php 8.0.0
  */
 
-class_alias('\Elasticsearch\Endpoints\Nodes\HotThreads', '\Elasticsearch\Endpoints\Cluster\Nodes\HotThreads');
-class_alias('\Elasticsearch\Endpoints\Nodes\Info', '\Elasticsearch\Endpoints\Cluster\Nodes\Info');
-class_alias('\Elasticsearch\Endpoints\Nodes\ReloadSecureSettings', '\Elasticsearch\Endpoints\Cluster\Nodes\ReloadSecureSettings');
-class_alias('\Elasticsearch\Endpoints\Nodes\Stats', '\Elasticsearch\Endpoints\Cluster\Nodes\Stats');
-class_alias('\Elasticsearch\Endpoints\Nodes\Usage', '\Elasticsearch\Endpoints\Cluster\Nodes\Usage');
-class_alias('\Elasticsearch\Endpoints\Cluster\GetSettings', '\Elasticsearch\Endpoints\Cluster\Settings\Get');
-class_alias('\Elasticsearch\Endpoints\Cluster\PutSettings', '\Elasticsearch\Endpoints\Cluster\Settings\Put');
-class_alias('\Elasticsearch\Endpoints\Indices\DeleteAlias', '\Elasticsearch\Endpoints\Indices\Alias\Delete');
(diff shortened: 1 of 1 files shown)


ezimuel:
@annuh thanks for this PR. I have only one question, why is `false` the autoloading in `class_exists()`?

annuh:
> @annuh thanks for this PR. I have only one question, why is `false` the autoloading in `class_exists()`?

I though `$autoload = false` would be a bit faster, but it should in both ways. See also: https://www.php.net/manual/en/function.class-exists.php#113852

Do you prefer this check without the `$autoload = true` option?

ezimuel:
If #1042 issue is related to the preloading feature of PHP I think we can use `false` since it means the alias has alrean been loaded during preload. I would like to test it before merge. I asked for more information about the preload configuration here https://github.com/elastic/elasticsearch-php/issues/1042#issuecomment-659255341.
@annuh can you test if it works with `true` and `false` as second parameter of `class_exists()`? Thanks.
I'll also try to build a test case for that.

ezimuel:
@annuh I did some test using the `opcache.preload` feature with `true` and `false`. It seems ok in both cases. We can use `false`, as you suggested. Thanks for the PR. I'll merge soon.

ezimuel:
Just released `elasticsearch-php` [7.9.0](https://github.com/elastic/elasticsearch-php/releases/tag/v7.9.0) with #1051 included

PR #1055, opened 2020-08-06
Show generic error messages when server returns no response

mlambley:
Although I just noticed you might want me to target an earlier branch. `6.x` perhaps? Let me know if you wish me to remake this PR.

### DIFF
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -613,12 +613,10 @@ class Connection implements ConnectionInterface
             return null;
         }
 
-        // if responseBody is not string, we convert it so it can be used as Exception message
-        if (!is_string($responseBody)) {
-            $responseBody = json_encode($responseBody);
-        }
-
-        if ($statusCode === 403) {
+        if (empty($responseBody)) {
+            //Provide a generic error message when the server returns no response
+            $exception = new BadRequest400Exception("Unknown {$statusCode} error from elasticsearch server", $statusCode);
+        } elseif ($statusCode === 403) {
             $exception = new Forbidden403Exception($responseBody, $statusCode);
         } elseif ($statusCode === 404) {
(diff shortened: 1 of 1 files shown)


src/Elasticsearch/Connections/Connection.php
```diff
@@ -613,10 +613,14 @@ private function process4xxError(array $request, array $response, array $ignore)
             return null;
         }
 
+        // if responseBody is not string, we convert it so it can be used as Exception message
         if (empty($responseBody)) {
-            //Provide a generic error message when the server returns no response
-            $exception = new BadRequest400Exception("Unknown {$statusCode} error from elasticsearch server", $statusCode);
-        } elseif ($statusCode === 403) {
+            $responseBody = "Unknown {$statusCode} error from elasticsearch server " . $exception->getMessage();
```
ezimuel, line 618: I think we can move that part into a `private` function in order to convert `$responseBody` into a string, using the "Unknown message" as you proposed in case of empty (or null) value. I'm thinking to something like this:
```php
private function convertBodyToString(
    $body, 
    int $statusCode, 
    ElasticsearchException $exception
) : string {
    if (empty($body)) {
        return sprintf(
            "Unknown %d error from Elasticsearch %s",
            $statusCode,
            $exception->getMessage()
        );
    }
    // if body is not string, we convert it so it can be used as Exception message
    if (!is_string($body)) {
        return json_encode($body);
    }
    return $body;
}
```
We should apply this in `process4xxError` and `process5xxError` as well.
@mlambley can you apply this refactor and send the PR against `7.x` branch? Thanks!

mlambley: I remade the PR for 7.x 
Please see https://github.com/elastic/elasticsearch-php/pull/1056

mlambley:
Closed in favour of https://github.com/elastic/elasticsearch-php/pull/1056

PR #1056, opened 2020-08-20
Show generic error messages when server returns no response

mlambley:


ezimuel:
Thanks @mlambley for this PR and sorry for the delay. I only changed the `$exception` as `Exception` instead of `ElasticsearchException` in `Connection::convertBodyToString` function. This is required by static code analysis since the `ElasticsearchException` interface is empty and we used `getMessage()` as part of the `Throwable` interface.

PR #1065, opened 2020-09-22
Fix for #1064 reset custom headers

### LINKED ISSUE #1064: Custom headers added on a specific endpoint are not reset after the call
When you add custom headers on a specific endpoint the headers are not reset after the call.
For instance, calling 2 indexes endpoint with a custom header on the first affects also the second call.
```php
$params = [
    'index' => 'my_index',
    'client' => [
        'headers' => [
            'Foo' => ['Bar']
        ]
    ]
];

// Create my_index with the custom header Foo: Bar
$response = $client->indices()->create($params);

$params = [
    'index' => 'my_index2',
];
// Create my_index2 still with the custom header Foo: Bar (stored from the previous, bug!)
$response = $client->indices()->create($params);
```

ezimuel:
This PR fixes #1064 resetting the custom headers, passed as parameters, after an endpoint call.

### DIFF
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -200,8 +200,9 @@ class Connection implements ConnectionInterface
             $body = $this->serializer->serialize($body);
         }
 
+        $headers = $this->headers;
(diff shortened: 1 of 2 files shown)


philkra (approved):
LGTM

Issue #1071, opened 2020-10-14
Incorrect documentation in Ignoring exception section


retnek:
Hi,

I've found a minor typo in your documentation here:
https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/per_request_configuration.html#_ignoring_exceptions
I think there's no MissingDocument404Exception rather Missing404Exception.
You can check this in github also: https://github.com/elastic/elasticsearch-php/search?q=MissingDocument404Exception

Cheers!

ezimuel:
@retnek thanks for reporting this. May you want to send a PR for fixing the documentation page https://github.com/elastic/elasticsearch-php/blob/7.x/docs/per-request-configuration.asciidoc ?
Let me know, thanks!

PR #1073, opened 2020-10-19
#1071 proper exception class name

retnek:
#1071 proper class names


ezimuel:
Thanks @retnek for your contribution.

PR #1063, opened 2020-09-16
Introducing PHP8 compliancy

giovannialbero1992:

PR #1075, opened 2020-10-26
Replaced array_walk with array_map in Connection::getURI

pascaldevink:
In Connection::getURI, array_walk was used together with passing the values as references, to change the original array.
Passing values as references is error-prone and discouraged for quite some time.
Also, when using in conjunction with PHP 8.0, it will fail.

array_map can do the same thing as the original array_walk implementation, but without the downsides of having side effects and having to pass values as references.

### DIFF
--- a/tests/Elasticsearch/Tests/Connections/ConnectionTest.php
+++ b/tests/Elasticsearch/Tests/Connections/ConnectionTest.php
@@ -363,4 +363,67 @@ class ConnectionTest extends \PHPUnit\Framework\TestCase
         $headersAfter = $connection->getHeaders();
         $this->assertEquals($headersBefore, $headersAfter);
     }
+
+    /**
+     * Test if the x-elastic-client-meta header is sent if $params['client']['x-elastic-client-meta'] is true  
+     */
+    public function testElasticMetaClientHeaderIsSentWhenParameterIsTrue()
+    {
+        $params = [
+            'client' => [
+                'x-elastic-client-meta'=> true
+            ]
+        ];
+        $host = [
+            'host' => 'localhost'
+        ];
+
+        $connection = new Connection(
+            ClientBuilder::defaultHandler(),
+            $host,
+            $params,
+            $this->serializer,
+            $this->logger,
+            $this->trace
+        );
+        $result  = $connection->performRequest('GET', '/');
+        $request = $connection->getLastRequestInfo()['request'];
+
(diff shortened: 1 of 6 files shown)


mfn:
This came up in https://github.com/elastic/elasticsearch-php/pull/1063#issuecomment-727770331

I took the changes in the PR and applied them to our local source tree in CI and can confirm that they work as intended!

OskarStark (approved):

giovannialbero1992:
Hi @mfn, can we cherry-pick this into #1063 ?

mfn:
@giovannialbero1992 🤷‍♀️

I see what you mean, didn't realize that #1063 targets master and this PR targets `7.10`. Technically I don't see a reason not to.

However I'm not familiar with the process here, just a reported at this point :) I'm used to the multiple branch model that fixes will be applied to the lowest version (`7.10` here) and then merged upstream into master anyway? But I think I'm not qualified to answer this, sorry.

pascaldevink:
@deguif @giovannialbero1992 PHP 8 was released today, is there any update on the progress of this PR?

deguif:
@ezimuel this one should probably be merged to fully support PHP 8 , as it fails otherwise on PHP 8

tests/Elasticsearch/Tests/Connections/ConnectionTest.php
```diff
@@ -363,4 +363,36 @@ public function testHeaderClientParamIsResetAfterSent()
         $headersAfter = $connection->getHeaders();
         $this->assertEquals($headersBefore, $headersAfter);
     }
+
+        /**
```
OskarStark, line 367: Indentation is wrong

OskarStark (changes requested):

tests/Elasticsearch/Tests/Connections/ConnectionTest.php
pascaldevink: Fixed

deguif:
@pascaldevink it should probably target `7.x`, and would you able to rebase so that the test failure disappear as it seems it was fixed in this branch

pascaldevink:
@deguif @ezimuel rebased and updated the target against `7.x`. The tests are succeeding again as well

ezimuel:
@pascaldevink actually, it was enough to remove the `&$key` reference in the previous `array_walk` in order to be compatible with PHP 8 but your code was better, thanks!

PR #1089, opened 2020-12-15
Add the x-elastic-client-meta header

ezimuel:
This PR adds a `x-elastic-client-meta` header in the HTTP request to send version information about the PHP language, the client library, the cURL version, the usage of async, etc.
An example of `x-elastic-client-meta` is as follows:
```
x-elastic-client-meta: es=7.10.0,php=7.4.11,a=0,c=7.68.0
```
where `es` is Elasticsearch version, `php`is PHP version, `a=0` means the HTTP call was sync (1=async), `c` is the cURL version.

This header can be disabled using the function `ClientBuilder::setElasticMetaHeader(false)`, as follows:
```php
    ->setElasticMetaHeader(false)  // default is true
    ->build();
```

### DIFF
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -169,6 +169,11 @@ class Connection implements ConnectionInterface
             phpversion()
         )];
 
+        // Add x-elastic-client-meta header, if enabled
+        if (isset($connectionParams['client']['x-elastic-client-meta']) && $connectionParams['client']['x-elastic-client-meta']) {
+            $this->headers['x-elastic-client-meta'] = [$this->getElasticMetaHeader($connectionParams)];
+        }
+
         $host = $hostDetails['host'];
         $path = null;
         if (isset($hostDetails['path']) === true) {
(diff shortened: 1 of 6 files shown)


src/Elasticsearch/Connections/Connection.php
```diff
@@ -574,6 +579,26 @@ protected function getCurlRetryException(array $request, array $response): Elast
         return $exception;
     }
 
+    /**
+     * Get the x-elastic-client-meta header
+     */
+    private function getElasticMetaHeader(array $connectionParams): string
+    {
+        $clientMeta = sprintf(
+            "es=%s,php=%s,a=%d",
```
sethmlarson, line 588: Need to add `t=...` for PHP transport version is the same as client version I assume?

sethmlarson (changes requested):
Woo this looks awesome! :star_struck: One tiny comment for you :)

src/Elasticsearch/Connections/Connection.php
ezimuel: I didn't insert the transport `t` parameter because I read this in the specification: `"If there is no transport library separate from the client library then the key must be omitted"`.

philkra (approved):
LGTM

sethmlarson (approved):

src/Elasticsearch/Connections/Connection.php
ezimuel: I added the `t` parameter since the spec was outdated.

PR #1096, opened 2021-01-11
Fix StickyRoundRobinSelector

reinder83:
When using a `SniffingConnectionPool` is can happen that a connection drops, having the current position inside the `StickyRoundRobinSelector` point to an unexisting connection at that moment, which results in an PHP Notice: undefined offset.

I have added an additional check if the offset exists before checking if the connection is alive, so it continues selecting the next node.


reinder83:
I had signed the CLA, the bot seems to disagree

reinder83:
This PR is already open for a while, is anything gonna happen? I have signed the CLA (again, to be sure) but I think, the other side needs to approve it before it's fully signed.

reinder83:
@ezimuel any chance this can be merged into 8.x branch instead?

Issue #676, opened 2017-11-08
Cannot use object of type Elasticsearch\Common\Exceptions\Missing404Exception as array


suchman:
Hi, first of all thanks for awesome library! But I found problem after updating on v5.3.1. When I try to check if document exists, error "Cannot use object of type Elasticsearch\Common\Exceptions\Missing404Exception as array" is shown on src\Elasticsearch\Transport.php row 118. $response varible contains object not an array ... 
Is it bug or Am I doing something wrong?

Thanks for reply!

polyfractal:
Argh, this is a bug.  I made a change in master which works on PHP 7 but not PHP 5.6, then backported the change to 5.x

I'll fix and push/tag in a minute, sorry about that!

polyfractal:
Tagged `v5.3.2`, which includes the previously committed (but not tagged) fix: https://github.com/elastic/elasticsearch-php/commit/c5db567d5a843a2c9778c7b74863357879abd34f

Thanks for the bug report!

suchman:
Works perfectly!
Thank you for quick fix :)

jkocour:
Hi, can you please apply the bugfix to v6.0?

PR #1109, opened 2021-03-02
Transport.php fix backported into 6.8.x

potemkin-git:
Following to [issue 676](https://github.com/elastic/elasticsearch-php/issues/676), fix is still missing from 6.8.x release

Error `Cannot use object of type Elasticsearch\Common\Exceptions\Missing404Exception as array` is still encountered in 6.8.

This PR intend to backport Master's fix into previous release 6.7.x and 6.8.x


ezimuel:
@potemkin-git thanks for your PR. I added the unit test from https://github.com/elastic/elasticsearch-php/pull/1066

Issue #989, opened 2020-01-17
ConnectionInterface::performRequest method signature does not match Connection::performRequest


nateeklund:
### Summary of problem or feature request
ConnectionInterface::performRequest method signature does not match Connection::performRequest method signature.

This causes mocking the class with Phake to fail.

### Code snippet of problem
ConnectionInterface.php
```php
    /**
     * @param  null $body
     * @return mixed
     */
    public function performRequest(string $method, string $uri, array $params = [], $body = null, array $options = [], Transport $transport = null);
```

Connection.php
```php
    /**
     * @param  string    $method
     * @param  string    $uri
     * @param  array     $params
     * @param  null      $body
     * @param  array     $options
     * @param  Transport $transport
     * @return mixed
     */
    public function performRequest(string $method, string $uri, ?array $params = [], $body = null, array $options = [], Transport $transport = null)
    {
       //  ....
    }
```

### System details

- PHP Version: 7.1
- ES-PHP client version: 7.4 and 7.5

nateeklund:
I'm happy to push up the fix for this, though if you'd rather just do it as part of whatever you're in the middle of that might make more sense as it's a 1 character change.
If I were to push it up I'd also fix up the PHPDoc comment a bit.

mamsincl:
https://github.com/elastic/elasticsearch-php/pull/1121 for version 6.8.0 - function signature missing all type hinting

PR #1121, opened 2021-03-26
Update Connection.php

mamsincl:
Fix function signature of performRequest to match declaration in ConnectionInterface

Issue:
```
Fatal error: 
Declaration of Elasticsearch\Connections\Connection::performRequest($method, $uri, $params = NULL, $body = NULL, $options = Array, ?Elasticsearch\Transport $transport = NULL) 
must be compatible with 
Elasticsearch\Connections\ConnectionInterface::performRequest(string $method, string $uri, ?array $params = Array, $body = NULL, array $options = Array, ?Elasticsearch\Transport $transport = NULL) 
in /mnt/volume-lon1-01/var/www/xxxx/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Connections/Connection.php on line 49
```

https://github.com/elastic/elasticsearch-php/issues/989 - but for 6.8.x function missing all type hinting

### DIFF
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -174,7 +174,7 @@ class Connection implements ConnectionInterface
      * @param \Elasticsearch\Transport $transport
      * @return mixed
      */
-    public function performRequest($method, $uri, $params = null, $body = null, $options = [], Transport $transport = null)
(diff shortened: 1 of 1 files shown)


ezimuel:
We are not supporting anymore 6.x, sorry for that.

Issue #1044, opened 2020-06-23
Scroll search with long scroll_id


c-leroy:
Reference documentation from https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/search_operations.html tells us to use scroll_id parameter in scroll() parameters. This leads to a Elasticsearch error in case scroll_id is too large, since - in this case - it is sent in request URI, which overcomes elasticsearch limits.

Example from https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/search_operations.html is :

```php
$client = ClientBuilder::create()->build();
$params = [
    'scroll' => '30s',          // how long between scroll requests. should be small!
    'size'   => 50,             // how many results *per shard* you want back
    'index'  => 'my_index',
    'body'   => [
        'query' => [
            'match_all' => new \stdClass()
        ]
    ]
];

// Execute the search
// The response will contain the first batch of documents
// and a scroll_id
$response = $client->search($params);

// Now we loop until the scroll "cursors" are exhausted
while (isset($response['hits']['hits']) && count($response['hits']['hits']) > 0) {

    // **
    // Do your work here, on the $response['hits']['hits'] array
    // **

    // When done, get the new scroll_id
    // You must always refresh your _scroll_id!  It can change sometimes
    $scroll_id = $response['_scroll_id'];

    // Execute a Scroll request and repeat
    $response = $client->scroll([
            'scroll_id' => $scroll_id,  //...using our previously obtained _scroll_id
            'scroll'    => '30s'        // and the same timeout window
        ]
    );
}
```
After viewing src/Elasticsearch/Endpoints/Scroll.php, we can see that : 
1) there is a warning about using scroll_id main parameter, that we didn't see (we are using kubernetes, maybe is it misconfigured?)
2) after a little reading inside, we guessed we could use the "body" parameter which would then end up with a sane URI using POST method and a body containing the needed scroll_id. It worked.

Correct form of scroll() with current version to avoid URI overflow : 

```php
      $response = $client->scroll([
        'body' => [
          'scroll_id' => $scroll_id,
          'scroll' => '30s'
        ]
      ]);
```

The same goes for deletion, the only way to delete a large scroll_id is : 
```php
      $client->clearScroll([
        'body' => [
          'scroll_id' => $scroll_id
        ]
      ]);
```
We are using elasticsearch-php 7.4.1

Regards

ezimuel:
[...]

PR #1134, opened 2021-04-29
Update search iterators to send "scroll_id" inside the request body

bakurin:
Moving scroll_id into requires body #1044 brakes `SearchResponseIterator` by triggering deprecation error.

The goal of this PR is to change the iterator classes as described in the [doc](https://github.com/elastic/elasticsearch-php/blob/master/docs/search-operations.asciidoc#scrolling) to avoid deprecation error

### DIFF
--- a/src/Elasticsearch/Helper/Iterators/SearchHitIterator.php
+++ b/src/Elasticsearch/Helper/Iterators/SearchHitIterator.php
@@ -76,8 +76,8 @@ class SearchHitIterator implements Iterator, \Countable
         }
 
         $this->count = 0;
-        if (isset($current_page['hits']) && isset($current_page['hits']['hits'])) {
-            $this->count = count($current_page['hits']['hits']);
+        if (isset($current_page['hits']) && isset($current_page['hits']['total'])) {
+            $this->count = $current_page['hits']['total']['value'] ?? $current_page['hits']['total'];
         }
 
         $this->readPageData();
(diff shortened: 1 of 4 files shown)


src/Elasticsearch/Helper/Iterators/SearchHitIterator.php
```diff
@@ -76,8 +76,8 @@ public function rewind(): void
         }
 
         $this->count = 0;
-        if (isset($current_page['hits']) && isset($current_page['hits']['total'])) {
-            $this->count = $current_page['hits']['total'];
+        if (isset($current_page['hits']) && isset($current_page['hits']['hits'])) {
+            $this->count = count($current_page['hits']['hits']);
```
ezimuel, line 80: We can use the `$current_page['hits']['total']` value of the result instead of `count()`. We need to check if `total` is a number or an array, since you can switch the behaviour using [rest_total_hits_as_hint](https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking-changes-7.0.html#hits-total-now-object-search-response).
We can use the following code:
```php
if (isset($current_page['hits']) && isset($current_page['hits']['total'])) {
    $this->count = $current_page['hits']['total']['value'] ?? $current_page['hits']['total'];
}
```

ezimuel (changes requested):
Thanks for this PR, it looks good. I only requested a little change about the usage of `count()`. Let me know what do you think. Thanks!

src/Elasticsearch/Helper/Iterators/SearchHitIterator.php
bakurin: Looks totally ok for me. I'll update the code.

ezimuel (approved):

ezimuel:
Thanks @bakurin, this PR will be released in `7.14.0`.

PR #1142, opened 2021-05-17
Add support for API Compatibility Header

ezimuel:
This PR adds the support for API compatibility header. Starting from version 7.13, Elasticsearch supports a compatibility header in `Content-Type` and `Accept`. The `elasticsearch-php` client can be configured to emit the following HTTP headers:
```
Content-Type: application/vnd.elasticsearch+json; compatible-with=7
Accept: application/vnd.elasticsearch+json; compatible-with=7
```
which signals to Elasticsearch that the client is requesting `7.x` version of request and response bodies. This allows forupgrading from 7.x to 8.x version of Elasticsearch without upgrading everything at once. Elasticsearch should be upgraded first after the compatibility header is configured and clients should be upgraded second.

To enable this compatibility header you need to create a `ELASTIC_CLIENT_APIVERSIONING` environment variable and set it to `true` or `1`.

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -635,11 +635,20 @@ class ClientBuilder
             if (! isset($this->connectionParams['client']['headers'])) {
                 $this->connectionParams['client']['headers'] = [];
             }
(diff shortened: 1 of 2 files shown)


philkra (approved):
LGTM

PR #1144, opened 2021-05-18
6.7.x

fghamsary:
This is just small changes in the code to be compatible with PHP 8 in composer.json and another deprecated usage of parameter after another parameter with default value.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     }
   ],
   "require": {
-    "php": "^7.0",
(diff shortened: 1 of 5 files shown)


derrabus:
I think, version 6.7 is not maintained anymore. Upgrading to 6.8 is not an option for you?

ezimuel:
@fghamsary I'm sorry but 6.7 is not supported anymore. You can upgrade to 6.8 that is PHP 8 compatible.

PR #1147, opened 2021-06-15
lower minimum php requirement

kylerberry:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     }
   ],
   "require": {
-    "php": "^7.3 || ^8.0",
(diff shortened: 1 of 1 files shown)


ezimuel:
@kylerberry why did you send (two times) this PR to bump min PHP requirement? Just curios.

kylerberry:
@ezimuel sorry, i was mistakenly targeting the wrong remote (source branch instead of my fork) - I'm doing a migration on a project from php7.2 to php8.0 but the es branch that introduces php8 (v6.8) supports minimum 7.3

I was forking to make some short term changes so that my build would pass.

very sorry

PR #1150, opened 2021-06-30
Refactor for abstract return type

bluedreamer:
PHP 7.4 complains now about return types on abstract functions. This is an issue for my using CirrusSearch mediawiki module. I see this is fixed in later branches but not for the 6.7.x stream

### DIFF
--- a/src/Elasticsearch/Endpoints/AbstractEndpoint.php
+++ b/src/Elasticsearch/Endpoints/AbstractEndpoint.php
@@ -45,20 +45,20 @@ abstract class AbstractEndpoint
     /** @var  SerializerInterface */
     protected $serializer;
 
-    /**
-     * @return string[]
-     */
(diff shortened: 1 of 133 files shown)


bluedreamer:
I did sign one, does the signing take time to propagate?

Electronic Record and Signature Disclosure: 
 Accepted: 6/30/2021 11:26:41 PM
 ID: 6cf300e3-5b1d-4787-be78-2ae94885d39b

bluedreamer:
Any advice on how to get the CLA passed and my guess is the CI failures need to be fixed too - but I don't known what is wrong. The errors seems nothing to do with my change (but must be) - any help apprecated as to where to look next.

ezimuel:
@bluedreamer sorry for the very late reply. Is this still an issue? If so, please send a new PR to `6.8.x` branch and I'll review asap. Thanks!

PR #1154, opened 2021-07-18
Allow psr/log 2

derrabus:
This PR enables this library to be used with the updated `psr/log` interfaces. The current code is already compatible with version 2 of the interfaces.

There's also a version 3, but unfortunately it requires a return type to be added to `Elasticsearch\Common\EmptyLogger`. Adding that type could break userland code that extends `EmptyLogger`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "php": "^7.3 || ^8.0",
     "ext-json": ">=1.3.7",
     "ezimuel/ringphp": "^1.1.2",
-    "psr/log": "~1.0"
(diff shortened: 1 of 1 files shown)


Jean85:
Ping @ezimuel: this is a blocker for https://github.com/Seldaek/monolog/pull/1564

ezimuel:
Thanks @derrabus for this PR and thanks @Jean85 for the ping. I'll merge it also in `7.x` and release in the upcoming `7.14.0` version.

PR #1158, opened 2021-08-25
Improve make.sh assemble Task

philkra:
* slightly refactored the `make.sh assemble <VERSION>` task
* added GitHub action to run the make assemble on each PR to ensure we can produce a sane artefact


Barberj859 (commented):
> #1161 #1159

PR #1159, opened 2021-08-26
Allow php 8.0+

xvilo:
I would like to use the 5.x version with PHP 8. It already ran and installed fine with `"php": "^7.4 || ^8.0"`. However upon removing `^7.4 || ` I can not install this package version anymore. 

This intends fixing it

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
     }
   ],
   "require": {
-    "php": "^5.6|^7.0",
+    "php": "^5.6|^7.0|^8.0",
     "psr/log": "~1.0",
     "guzzlehttp/ringphp" : "~1.0"
   },


mhkhung:
At least 2 places in the code needs to be fixed:

```
RuntimeException: (errorno: 8192) Required parameter $transport follows optional parameter $params on line 98 in file /srv/work/Sprout/application/libraries/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Connections/ConnectionInterface.php

RuntimeException: (errorno: 8192) Required parameter $connectionPool follows optional parameter $sniffOnStart on line 51 in file /srv/work/Sprout/application/libraries/vendor/elasticsearch/elasticsearch/src/Elasticsearch/Transport.php
```

ezimuel:
@xvilo I'm sorry but we are not supporting anymore `5.x`. We still support `6.8.x` as minimum version.

PR #1163, opened 2021-09-13
[DOCS] Added compatibility section in doc + updated release notes

ezimuel:
This PR provides the following changes in the documentation:

- added a compatibility section;
- updated the release notes up to 7.14;

Moreover it updates the README with the new compatibility section content.


docs/compatibility.asciidoc
```diff
@@ -0,0 +1,7 @@
+[[compatibility]]
+== Compatibility
+
+Language clients are forward compatible; meaning that clients support communicating
+with greater or equal minor versions of Elasticsearch. Elasticsearch language
+clients are only backwards compatible with default distributions and without
+guarantees made.
```
szabosteve, line 7: In the rest of the Client books, we added the compatibility information to the `Installation` section as a level-three sub-section:
```
[discrete]
[[compatibility]]
=== Compatibility
```
If there is no particular reason to add this info as a separate section, I would suggest moving this content under the Installation section.

docs/index.asciidoc
```diff
@@ -5,6 +5,8 @@ include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 
 include::overview.asciidoc[]
 
+include::compatibility.asciidoc[]
```
szabosteve, line 8: If you move the compatibility info under `Installation`, then this item can be deleted.
```suggestion
```

szabosteve (commented):
One suggestion about the place of the compatibility info inside the PHP book.

ezimuel (commented):
I just moved the `compatibility` section into `installation` as suggested.

szabosteve (approved):
LGTM!

PR #1178, opened 2021-11-23
Fix #1176 rename includePortInHostHeader to setIncludePortInHostHeader

### LINKED ISSUE #1176: includePortInHostHeader not usable via fromConfig

The `ClientBuilder` `includePortInHostHeader()` method is not accessible when using `fromConfig()` as only `set`-prefixed methods are applicable in `fromConfig`.

As I am using a wrapper package for Laravel Scout that uses the `fromConfig` method, I can't initiate the Client myself so it's a bit of an issue.

huanes:
This PR modifies #997 by renaming the added method and therefore fixes #1176 as the aforementioned method is now accessible via `ClientBuilder` `fromConfig` as well.

### DIFF
--- a/docs/http-meta-data.asciidoc
+++ b/docs/http-meta-data.asciidoc
@@ -39,7 +39,7 @@ as follows:
 [source,php]
 ----
 $client = Elasticsearch\ClientBuilder::create()
-    ->includePortInHostHeader(true)
+    ->setIncludePortInHostHeader(true)
     ->build();
 ----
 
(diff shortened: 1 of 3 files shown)


ezimuel:
@huanes thanks for the PR. Unfortunately, we cannot rename `includePortInHostHeader` in `setIncludePortInHostHeader` because this will be a BC break. I proposed a different approach in #1181 to support  `includePortInHostHeader` in ClientBuilder::fromConfig.

ezimuel:
Closing in favor or #1181

Issue #1171, opened 2021-10-15
Error encoded data query decoding problem


vst93:
### Summary of problem or feature request

Similar to this data `ud83d\ude4f`, the interception of the error before being stored, only the `\UD83D` section, but PHP's JSON_Decode cannot correctly resolve this data, which will cause Search to throw a fatal error, this You can only give up all data (there may be a lot of queries at a time), or you can only report the error.
But I need a batch export, I can't give up other data, so I can only modify the method in the official package.

vendor/elasticsearch/elasticsearch/src/Elasticsearch/Serializers/SmartSerializer.php
![image](https://user-images.githubusercontent.com/9971164/137427020-560c4361-84db-42e7-bd0a-0de9e1a99edf.png)
![image](https://user-images.githubusercontent.com/9971164/137427099-cebc8a3c-f18a-4b13-bb11-41a3177269f1.png)

##### This is temporarily solved in my problem, but it is still hoped that the official can give a solution or suggestion.
I know that the wrong data must be able to correctly analyze some unreasonable, but the work of multiplayer games is helpless.😮‍💨

The problem is similar to this：[https://stackoverflow.com/questions/32799627/elasticsearch-field-showing-as-bad-string-causes-error-when-searched](https://stackoverflow.com/questions/32799627/elasticsearch-field-showing-as-bad-string-causes-error-when-searched)

### Code snippet of problem

```php

    if (version_compare(PHP_VERSION, '7.3.0') >= 0) {
            try {
                $result = json_decode($data, true, 512, JSON_THROW_ON_ERROR);
                return $result;
            } catch (\JsonException $e) {
                //遇到反斜杠解析失败的尝试处理
                try {
                    $data = str_replace("\\",'\\\\',$data);
                    $result = json_decode($data, true, 512, JSON_THROW_ON_ERROR);
                    return $result;
                }catch (\JsonException $e){
                    $result = $result ?? [];
                    throw new JsonErrorException($e->getMessage(), $data, $result);
                }
//                $result = $result ?? [];
//                throw new JsonErrorException($e->getMessage(), $data, $result);
            }
        }

```

### System details

- Operating System : macos
- PHP Version : 7.3
- ES-PHP client version : 7.3
- Elasticsearch version : 7.3

ezimuel:
[...]

PR #1179, opened 2021-11-23
Fix for #1171 utf-16 issue in SmartSerializer

ezimuel:
This PR contains the fix for #1171 related to single unpaired UTF-16 surrogate in unicode escape. In case of `JSON_ERROR_UTF16` the `SmartSerilizer::decode()` function escape the backslash `\` using the double backslash `\\`.


technige (approved):
LGTM :+1:

PR #1181, opened 2021-12-06
Fix #1176 adding includePortInHostHeader in fromConfig

### LINKED ISSUE #1176: includePortInHostHeader not usable via fromConfig

ezimuel:
This PR fixes #1176 adding includePortInHostHeader support in ClientBuilder::fromConfig.

### DIFF
--- a/src/Elasticsearch/ClientBuilder.php
+++ b/src/Elasticsearch/ClientBuilder.php
@@ -197,7 +197,8 @@ class ClientBuilder
     {
         $builder = new static;
         foreach ($config as $key => $value) {
-            $method = "set$key";
+            $allowedMethods = ['includePortInHostHeader']; // other available methods
+            $method = in_array($key, $allowedMethods) ? $key : "set$key";
             $reflection = new ReflectionClass($builder);
             if ($reflection->hasMethod($method)) {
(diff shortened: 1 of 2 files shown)


src/Elasticsearch/ClientBuilder.php
```diff
@@ -197,7 +197,8 @@ public static function fromConfig(array $config, bool $quiet = false): Client
     {
         $builder = new static;
         foreach ($config as $key => $value) {
-            $method = "set$key";
+            $allowedMethods = ['includePortInHostHeader']; // other available methods
```
technige, line 200: Might it be better to move this hard-coded list of allowed methods up to class scope? It would be potentially easier to find there, and wouldn't get created for each loop iteration.

ezimuel: Good suggestion, I'll move it using a class constant.

technige (approved):
Perfetto!

PR #1184, opened 2021-12-15
Allow psr/log v3

ruudk:

PR #1187, opened 2021-12-21
Support `psr/log:^3`

Ocramius:
See https://github.com/php-fig/log/compare/2.0.0...3.0.0

`psr/log:3.0.0` adds return types to all signatures: we don't implement `psr/log`, but only consume it, so we should be fine here.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "php": "^7.3 || ^8.0",
     "ext-json": ">=1.3.7",
     "ezimuel/ringphp": "^1.1.2",
-    "psr/log": "^1|^2"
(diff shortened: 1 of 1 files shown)


Ocramius:
It's signed, but the CLA process verification craps out because of blacklisted domains that do network tracking, and are (rightfully) blocked by my own network rules.

Feel free to look it up.

ruudk:
Duplicate of https://github.com/elastic/elasticsearch-php/pull/1184

technige:
Closing, as duplicate of #1184. Please check for existing PRs before submitting a new one.

PR #1189, opened 2021-12-29
Remove IF statement

cronlabspl:


ezimuel:
@cronlabspl actually we need the if inside the foreach beacuse the value can be a subarray.

PR #1190, opened 2021-12-31
Allow psr/log 3

franmomu:
There is https://github.com/elastic/elasticsearch-php/pull/1184 and https://github.com/elastic/elasticsearch-php/pull/1187, but it needs a little bit more of work since there are implementations of psr/log.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "php": "^7.3 || ^8.0",
     "ext-json": ">=1.3.7",
     "ezimuel/ringphp": "^1.1.2",
-    "psr/log": "^1|^2"
+    "psr/log": "^1|^2|^3"
   },
   "require-dev": {
     "ext-yaml": "*",
--- a/src/Elasticsearch/Common/EmptyLogger.php
+++ b/src/Elasticsearch/Common/EmptyLogger.php
@@ -32,7 +32,7 @@ class EmptyLogger extends AbstractLogger implements LoggerInterface
     /**
      * {@inheritDoc}
      */
-    public function log($level, $message, array $context = [])
+    public function log($level, $message, array $context = []): void
     {
         return;
     }
--- a/tests/Elasticsearch/Tests/ClientBuilder/ArrayLogger.php
+++ b/tests/Elasticsearch/Tests/ClientBuilder/ArrayLogger.php
@@ -17,57 +17,16 @@ declare(strict_types = 1);
 
 namespace Elasticsearch\Tests\ClientBuilder;
 
-use Psr\Log\LoggerInterface;
-use Psr\Log\LogLevel;
+use Psr\Log\AbstractLogger;
 
-class ArrayLogger implements LoggerInterface
+class ArrayLogger extends AbstractLogger
 {
     /**
      * @var mixed[]
      */
     public $output = [];
 
-    public function emergency($message, array $context = array())
-    {
-        $this->log(LogLevel::EMERGENCY, $message, $context);
-    }
-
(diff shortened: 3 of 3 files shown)


src/Elasticsearch/Common/EmptyLogger.php
```diff
@@ -32,7 +32,7 @@ class EmptyLogger extends AbstractLogger implements LoggerInterface
     /**
      * {@inheritDoc}
      */
-    public function log($level, $message, array $context = [])
+    public function log($level, $message, array $context = []): void
```
franmomu, line 35: note that this is BC break, but given the name of the class, I guess it is not meant to be extended.

composer.json
```diff
@@ -19,7 +19,7 @@
     "php": "^7.3 || ^8.0",
     "ext-json": ">=1.3.7",
     "ezimuel/ringphp": "^1.1.2",
-    "psr/log": "^1|^2"
+    "psr/log": "^1|^2|^3"
```
cronlabspl, line 22: PSR-2 is [deprecated](https://www.php-fig.org/psr/psr-2/) in favour of [PSR-12](https://www.php-fig.org/psr/psr-12/)

cronlabspl (changes requested):
PSR-2 is [deprecated](https://www.php-fig.org/psr/psr-2/) in favour of [PSR-12](https://www.php-fig.org/psr/psr-12/)

composer.json
franmomu: I think you're mixing things here, [`psr/log`](https://github.com/php-fig/log) package is about [`PSR-3`](https://www.php-fig.org/psr/psr-3/) which currently has 3 versions.

PSR-2 and PSR-12 are about coding standard

cronlabspl: @franmomu correct, but it doesnt change the fact that PSR-2 is deprecated

franmomu: so... what should I change in this PR?

cronlabspl: remove `^2` from line 22 in `composer.json`

franmomu: Please read this again:

> I think you're mixing things here, [`psr/log`](https://github.com/php-fig/log) package is about [`PSR-3`](https://www.php-fig.org/psr/psr-3/) which currently has 3 versions.

`psr/log` has nothing to do with psr-2

cronlabspl: > `psr/log` has nothing to do with psr-2

Yes. So will close this as misunderstanding. Must have been.

ruudk:
@cronlabspl Can you please remove or hide your comment as it's confusing (see thread).

@ezimuel @karyna-tsymbal-atwix @jrodewig Can this PR be merged? Multiple people are now proposing the same change. This change is needed to be able to upgrade to Symfony 6.

technige:
This is a duplicate of #1184 and #1187, the latter of which is now closed as an exact duplicate.

@franmomu We very much appreciate the extra detail you have added to this issue, but please could you instead collaborate on #1184 (the original issue), rather than opening a new one. This will help to expedite our work in reviewing and merging. Once you have do so, this issue can be closed.

franmomu:
@technige no problem merging that one!

I created this one because I was able to run the workflow jobs and the other PRs I thought they were created editing the file from Github directly, so the purpose was in fact to ease the review (avoid approving the workflow, see if the tests passes, make changes, approve again, etc), but as I said no worries at all merging the other one.

ruudk:
This can be closed as it has been merged into https://github.com/elastic/elasticsearch-php/pull/1184

PR #1191, opened 2022-01-04
remove PSR-2 || Introduce PSR-12

cronlabspl:
PSR-2 is [deprecated](https://www.php-fig.org/psr/psr-2/) in favour of [PSR-12](https://www.php-fig.org/psr/psr-12/)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
     "php": "^7.3 || ^8.0",
     "ext-json": ">=1.3.7",
     "ezimuel/ringphp": "^1.1.2",
-    "psr/log": "^1|^2"
+    "psr/log": "^1|^3"
   },
   "require-dev": {
     "ext-yaml": "*",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
     "php": "^7.3 || ^8.0",
     "ext-json": ">=1.3.7",
     "ezimuel/ringphp": "^1.1.2",
-    "psr/log": "^1|^2"
+    "psr/log": "^1|^3"
```
ruudk, line 19: this has nothing todo with the coding standard, please revisit your PR.

cronlabspl: `PSR`'s are no coding standards? Really?

ruudk: psr-2 and psr-12 yes, but `psr/log` is https://www.php-fig.org/psr/psr-3/ , a different standard

ruudk: :facepal

ezimuel:
@cronlabspl as suggested also by @ruudk please remove the `psr/log` change and leave only the `CONTRIBUTION.md` changes. Thanks!

PR #1192, opened 2022-01-04
General fixes

cronlabspl:


ezimuel:
@cronlabspl thanks for this PR but it includes some changes that we want to preserve. We need the preserver the `util/template` folder as is because this is used to generate the endpoint code and the YAML tests for the library. Moreover, we already bump the tests with PHP 8.1.

Issue #918, opened 2019-07-15
How to set connection pool if the es nodes is behind the load balance in PHP?


guangfanqu:
### Summary of problem or feature request

I have a elasticsearch cluster, the cluster is behind the load balance. 
And the network client to cluster node is unreachable.

If set default pool `staticNoPingConnectionPool` and if one of the nodes is dead then the client will response "No alive nodes found in your cluster" in several minutes even if the dead node had returned to normal state .

In the article "https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/connection_pool.html" I don't find the best solution. I even think the `SimpleConnectionPool` is better than other solutions.

So who can give me some suggestion?

Thanks.

### System details

- Operating System
Linux
- PHP Version
7.1.25
- ES-PHP client version
6.7.0
- Elasticsearch version
6.7.0

ezimuel:
@guangfanqu if I got your point, the load balancer is the only service that you can connect from `elasticsearch-php`, right? If this is the case, you will have a domain name (or an IP address) for your Elasticsearch cluster managed by the load balancer. In your scenario, you don't need a connection pool from the client because it's managed by the load balancer.

In this case, I suggest to have a look at the [simpleConnectionPool](https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/connection_pool.html#_simpleconnectionpool) using only the domain service of Elasticsearch (or IP address) managed by the load balancer.

```php
$client = ClientBuilder::create()
    ->setConnectionPool('\Elasticsearch\ConnectionPool\SimpleConnectionPool', [])
    ->setHosts(['myelasticsearch.domain:9200'])
    ->build();
```

guangfanqu:
@ezimuel 

Thanks a lot.

PR #1193, opened 2022-02-01
[docs] Recommend simpleConnectionPool for Cloud

toby-sutor:
As per https://github.com/elastic/elasticsearch-php/issues/918#issuecomment-511337920 and https://elastic.slack.com/archives/CH37CF39S/p1643707941052409 we recommend using the `simpleConnectionPool` when there is a proxy or load balancer in place. This applies to Elasticsearch Service and ECE (Cloud). Hence the comments need to be updated.


ezimuel:
Thanks @toby-sutor

PR #1205, opened 2022-03-28
Allow `psr/log` v3

ruudk:


ruudk:
I was too fast, turns out we need `psr/log` v3 in [elastic-transport](https://github.com/elastic/elastic-transport-php/pull/6)

PR #1211, opened 2022-03-31
bugfix ClientEndpointsTrait

amirfaramarzi:
In the check Required Parameters function, the body is checked (first line)

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -164,8 +164,6 @@ final class Client
         $response = $this->transport->sendRequest($request);
         $this->logger->info(sprintf("Response time in %.3f sec", microtime(true) - $start));       
 
-        $result = new Elasticsearch;
-        $result->setResponse($response, $request->getMethod() === 'HEAD' ? false : $this->getResponseException());
-        return $result;
+        return (new Elasticsearch())->setResponse($response, $request->getMethod() === 'HEAD' ? false : $this->getResponseException());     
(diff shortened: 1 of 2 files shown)


src/Client.php
```diff
@@ -164,8 +164,6 @@ public function sendRequest(RequestInterface $request)
         $response = $this->transport->sendRequest($request);
         $this->logger->info(sprintf("Response time in %.3f sec", microtime(true) - $start));       
 
-        $result = new Elasticsearch;
-        $result->setResponse($response, $request->getMethod() === 'HEAD' ? false : $this->getResponseException());
-        return $result;
+        return (new Elasticsearch())->setResponse($response, $request->getMethod() === 'HEAD' ? false : $this->getResponseException());
```
ezimuel, line 167: This code is very difficult to read, I'll prefer to leave as is. Simple is beautiful :-)

ezimuel:
@amirfaramarzi the `ClientEndpointsTrait.php` is a generated file. You can see the `@generated` phpdoc at [the beginning](https://github.com/elastic/elasticsearch-php/blob/master/src/Traits/ClientEndpointsTrait.php#L27) of the Trait. That means we cannot change the file directly, we need to fix the code generator, that is an internal project at Elastic.
I'm closing this PR and I'll fix this in the code generator. Thanks for your contribution.

Issue #1218, opened 2022-04-25
Cannot deserialize the reponse as array with Content-Type: text/plain


stayallive:
### Summary of problem or feature request

When trying to get aliases from Elastic I get the following error:

```
Cannot deserialize the reponse as array with Content-Type: text/plain; charset=UTF-8
./vendor/elasticsearch/elasticsearch/src/Response/Elasticsearch.php:116 { …}
./vendor/elasticsearch/elasticsearch/src/Response/Elasticsearch.php:190 { …}
```

This worked on 7.x and I cannot figure out why it would fail now all of a sudden since I don't think this method is deprecated on 8.x or there is a better way to retrieve this. Probably missing something obvious so any help is appreciated!

### Code snippet of problem

```php
$aliases = $this->esClient->cat()->aliases([
    'name' => 'some-name',
]);

return $aliases[0] ?? null;
```

### System details

- Operating System: Alpine Linux 3.15
- PHP Version: PHP 7.4.29
- ES-PHP client version: 8.1.0
- Elasticsearch version: 8.1.3

ezimuel:
@stayallive just sent this PR https://github.com/elastic/elasticsearch-php/pull/1220. I'll release it with 8.2.0. Let me know if you have any feedback, thanks!

stayallive:
Looks good to me! 💪

ezimuel:
@stayallive I'm also thinking to enable by default the `Accept` header to `application/json` format instead of `text/plain` for the `cat()` endpoints. Using this approach, we can have the same behaviour of `7.x`.

stayallive:
I'm not too familiar with these API's but a plain text response seems... less useful to me in almost all cases. So this sounds very logical to me.

ezimuel:
@stayallive actually, I cannot change the `Accept` header of these endpoints since I'll change the official response from Elasticsearch. I think we will suggest to use the `"format" => "json"` parameter if someone needs to have array as response.

stayallive:
Ah right, that makes sense... for the next version maybe 👍 

Thanks again for the quick fix and workaround!

PR #1220, opened 2022-05-03
Added the array support for text/plain

ezimuel:
This PR adds the support of array in Elasticsearch response when the `Content-Type` is `text/plain`. This should solve #1218


picandocodigo (approved):

ezimuel:
Thanks @picandocodigo !

Issue #1221, opened 2022-05-03
Version 8 is missing the API Compatibility Header


ezimuel:
We need to re-apply yhe API Compatibility header, provided in https://github.com/elastic/elasticsearch-php/pull/1142 for `7.x` branch.

ezimuel:
Solved in https://github.com/elastic/elasticsearch-php/pull/1233

PR #1233, opened 2022-06-27
Add the API compatibility header in v8

ezimuel:
This PR adds the API compatibility header in client v8. This should solve issue https://github.com/elastic/elasticsearch-php/issues/1221

### DIFF
--- a/src/Traits/EndpointTrait.php
+++ b/src/Traits/EndpointTrait.php
@@ -14,6 +14,7 @@ declare(strict_types = 1);
 
 namespace Elastic\Elasticsearch\Traits;
 
+use Elastic\Elasticsearch\Client;
 use Elastic\Elasticsearch\Exception\ContentTypeException;
 use Elastic\Elasticsearch\Exception\MissingParameterException;
 use Elastic\Transport\Serializer\JsonSerializer;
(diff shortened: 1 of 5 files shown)


src/Traits/EndpointTrait.php
```diff
@@ -106,10 +107,12 @@ protected function addQueryString(string $url, array $params, array $keys): stri
      */
     protected function bodySerialize($body, string $contentType): string
     {
-        if (strpos($contentType, 'application/x-ndjson') !== false) {
+        if (strpos($contentType, 'application/x-ndjson') !== false ||
```
technige, line 110: Is there ever a positive case when `strpos(...)` wouldn't just be 0, rather than just non-false? It's a weird (almost non-existent) edge case, but `snapplication/x-ndjson` or something similar would get through this?

So is there any reason why we couldn't change `!== false` to `=== 0`, to tighten it up?

Alternatively, use `str_starts_with` to be even clearer.

technige (approved):
LGTM

src/Traits/EndpointTrait.php
ezimuel: Because `0` means `application/x-ndjson` is present in `$contentType` as the beginning of the string. Using `false` we are more safe, becuase we are sure that no substring is present.

NikosDevPhp:
@ezimuel Shouldn't this be optional, having a flag to control it?

We are currently trying to use client v.8 with server v.7.
I understand the compatibility header is for the other way around (working with server v.8 and client v.7) but I see requests are working okay in our setup without compatibility headers added.

Issue #1235, opened 2022-07-06
elasticsearch/elasticsearch 6.8.0 removes IndicesNamespace::deleteMapping


cweiske:
### Summary of problem or feature request

Version 6.8.0 removes the method `Elasticsearch\Namespaces\IndicesNamespace::deleteMapping()` with commit 2da5f9337a269113954a84fe8e79ea6843f0e7c3.

This method existed in 6.7.2.

Per semantic versioning, feature releases like the update from 6.7 to 6.8 may not break existing APIs, and here the API has been broken.
This leads to problems in the drupal module elasticsearch_connector: https://www.drupal.org/project/elasticsearch_connector/issues/2824539

ezimuel:
@cweiske this is a very old feature that [has been removed](https://github.com/elastic/elasticsearch/issues/8877) in Elasticsearch 2.0. Unfortunately, since v.6.8.0 we didn't have a code generator in place and this `Elasticsearch\Namespaces\IndicesNamespace::deleteMapping()` remained unsed since than. This means even if `elasticsearch-php` exposes it this do not have any effect on Elasticsearch :disappointed: . With `6.8.0` we released a code generator for mapping the Elasticsearch API specification. That's why this function has been removed automatically since it was not anymore part of the latest API.
We should have included this in the BREAKING_CHANGES.md file. I'm sorry for the inconvenient :cry:

PR #1238, opened 2022-07-14
Restore IndicesNamespace::deleteMapping() to keep backwards compatibility

cweiske:
With 6.8.0, this method got removed, which broke the semantic
versioning contract.

This commit restores the method so that applications using that method do
not break.

See https://www.drupal.org/project/elasticsearch_connector/issues/2824539

Resolves: https://github.com/elastic/elasticsearch-php/issues/1235


cweiske:
I will not sign a CLA.

PR #1249, opened 2022-08-23
Added a ClientInterface

### LINKED ISSUE #1227: Client should implement an interface to make unittesting possible.

Since client was made final we can not mock the class anymore. However, good practice would tell us that we need to mock the edges of our systems. Of course ES is such an edge. 

Therefore, whenever important classes like this are made final, an interface should be created to help us mock that instead. Of course it also helps us think about the system through an interface instead of an implementation. 

I assume there are more people running into this problem so I have a pull request ready to go.

ezimuel:
This PR fixes #1227

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -68,11 +68,17 @@ final class Client implements ClientInterface
         $this->defaultTransportSettings($this->transport);
     }
 
+    /**
+     * @inheritdoc
+     */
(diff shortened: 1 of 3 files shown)


picandocodigo (approved):

ezimuel:
@hkulekci good catch, thanks! I updated the PR.

PR #1253, opened 2022-09-22
Fix for StickyRoundRobinSelector starts with second host

### LINKED ISSUE #1242: StickyRoundRobinSelector starts with second host

When a list of multiple hosts is configured, `randomizeHosts` is disabled and `StickyRoundRobinSelector` is used the first connection goes to the second host (instead to the first one).

e.g. hosts:
- host1
- **host2**
- host3
- host4

Are tried in this order: **host2**, host3, host4, host1

### Code snippet of problem

From `StickyRoundRobinSelector::select()`:

        if ($connections[$this->current]->isAlive()) {
            return $connections[$this->current];
        }   

Because the first connection is not alive (yet), the first connection is skipped first time when `select()` is called.

### System details

elasticsearch-php v7.10.0

ezimuel:
This PR fixes #1242. We changed the `isAlive` default setting to `true` for the `Connection` class. In the past we  assumed all the hosts should be set dead at the beginning. We think this should not impact other Connection Pools, hopefully.

### DIFF
--- a/src/Elasticsearch/Connections/Connection.php
+++ b/src/Elasticsearch/Connections/Connection.php
@@ -99,7 +99,7 @@ class Connection implements ConnectionInterface
     /**
      * @var bool
      */
-    protected $isAlive = false;
+    protected $isAlive = true;
 
     /**
      * @var float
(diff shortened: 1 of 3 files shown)


picandocodigo (approved):

ezimuel:
Thanks @picandocodigo for the review!

PR #1255, opened 2022-09-27
Added CI make bump action

ezimuel:
As titled.

### DIFF
--- a/.ci/make.sh
+++ b/.ci/make.sh
@@ -154,16 +154,18 @@ fi
 
 if [[ "$CMD" == "bump" ]]; then
     # Change version to src/Client.php
-    sed -i "s/const VERSION = '[0-9]\+\.[0-9]\+\.[0-9]\+'/const VERSION = '$VERSION.0'/" $repo/src/Client.php
-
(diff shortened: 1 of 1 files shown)


.ci/make.sh
```diff
@@ -154,12 +154,16 @@ fi
 
 if [[ "$CMD" == "bump" ]]; then
     # Change version to src/Client.php
-    # sed -i "s/const VERSION = '[0-9]\+.[0-9]\+.[0-9]\+\(-dev\)\?'/const VERSION = '$VERSION'/g" $repo/src/Client.php 
+    sed -i "s/const VERSION = '[0-9]\+\.[0-9]\+\.[0-9]\+'/const VERSION = '$VERSION.0'/" $repo/src/Client.php
```
sethmlarson, line 157: I believe `$VERSION` is the complete three-number version (ie `8.3.4`). Might have to update below to match the expected input format.

ezimuel: I fixed this extracting a `$MINOR_VERSION` from `$VERSION`. Thanks!

PR #1263, opened 2022-11-01
Changed VERSION to 8.5.1

picandocodigo:

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -28,7 +28,7 @@ use Psr\Log\LoggerInterface;
 final class Client implements ClientInterface
 {
     const CLIENT_NAME = 'es';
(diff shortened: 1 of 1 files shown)


sethmlarson (commented):
From the [bump script](https://github.com/elastic/elasticsearch-php/blob/main/.ci/make.sh#L165), do we have to update the unified-release.yml too?

picandocodigo:
@sethmlarson I think since there's going to be a build of `8.5.1-SNAPSHOT`, updating the stack would mean the build would fail while the artifacts are not available?

sethmlarson:
@picandocodigo Should the bump script be updated then? I'll approve this one.

PR #1303, opened 2023-03-24
Added ELASTIC_CLIENT_URL_PLUS_AS_SPACE env variable for URL encode

### LINKED ISSUE #1278: Get/update fails for item IDs that contain spaces
### Summary
I want to update or get a record having id containing a space using the PHP client library.
Expected: The record having the correct id is returned/updated
Actual: The record is not found

### Code snippet of problem

**Test dataset:**

Kibana dev tools for inserts:

POST test_index/doc/_bulk
{"index": {"_index": "test_index", "_id": "test test"}}
{"field1": "value1"}

This will return data properly:
GET /test_index/_doc/test%20test

```
$params = [
  'index' => 'test_index',
  //'id'    => 'lp-3MIYB7NAUVXHg5pvt', THIS is working perfectly(no spaces)
  'id'    => 'test test', // this is not working because urlencode does not respect the url encode standard.
  'type' => '_doc'
];

$response = $elasticClient->get($params);
```

I think the problem is caused by the usage of urlencode instead of rawurlencode in the file AbstractEndPoint, not sure about how many client's methods are affected, not sure if the latest versions are affected, but I highly suspect they are.

See: https://www.php.net/manual/en/function.urldecode.php
See: https://www.php.net/manual/en/function.rawurlencode.php

### System details
- Ubuntu
- 7.3
- 6.x//7.x
- 7.17

ezimuel:
This PR should fix #1278 adding an env variable `ELASTIC_CLIENT_URL_PLUS_AS_SPACE` that can be used to enable the URL encoder for translate a space character with `%20` instead of `+`. The default value for the space character is `+`, to prevent BC break.

### DIFF
--- a/src/Elasticsearch/Utility.php
+++ b/src/Elasticsearch/Utility.php
@@ -18,7 +18,7 @@ namespace Elasticsearch;
 
 class Utility
 {
-    const ENV_URL_PLUS_AS_SPACE = 'ES_URL_PLUS_AS_SPACE';
+    const ENV_URL_PLUS_AS_SPACE = 'ELASTIC_CLIENT_URL_PLUS_AS_SPACE';
 
     /**
      * Get the ENV variable with a thread safe fallback criteria
--- a/src/Elasticsearch/Utility.php
+++ b/src/Elasticsearch/Utility.php
@@ -1,0 +1,47 @@
+<?php
+/**
+ * Elasticsearch PHP client
+ *
+ * @link      https://github.com/elastic/elasticsearch-php/
+ * @copyright Copyright (c) Elasticsearch B.V (https://www.elastic.co)
+ * @license   http://www.apache.org/licenses/LICENSE-2.0 Apache License, Version 2.0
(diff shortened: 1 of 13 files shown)


src/Elasticsearch/Utility.php
```diff
@@ -0,0 +1,47 @@
...
+ * the GNU Lesser General Public License, Version 2.1, at your option.
+ * See the LICENSE file in the project root for more information.
+ */
+declare(strict_types = 1);
+
+namespace Elasticsearch;
+
+class Utility
+{
+    const ENV_URL_PLUS_AS_SPACE = 'ES_URL_PLUS_AS_SPACE';
```
sethmlarson, line 21: We've used `ELASTIC_CLIENT_` as our prefix for env vars historically, what are your thoughts on using that prefix here ad well?

ezimuel: Sounds good!

PR #1339, opened 2023-07-17
[DOCS] Restructures repo readme

szabosteve:
## Overview

This PR adds links to the different sections of the README file that point to the corresponding section of the getting started page. This way only the getting started documentation needs to be maintained and it will be the source of truth.


ezimuel (approved):
LGTM!

Issue #1349, opened 2023-08-21
/docs shipped in prod package


dkarlovi:
### Summary of problem or feature request

`/docs` is included in the final package (as is PHPStan and PHPUnit config). The package is 4MB in size, with the docs being 1.2MB, making it 1/3 of the size, while it's very unlikely the docs will be read from the production install of the package, the same principle applies as for tests which are excluded, linking to the repo where the docs can be read should be enough for this purpose.

### System details

- ES-PHP client version 7.17.2 release package via Packagis

PR #1350, opened 2023-08-22
add more folders to .gitattributes

dmaicher:
Fixes https://github.com/elastic/elasticsearch-php/issues/1349


ezimuel:
Thanks @dmaicher and sorry for the late reply.

ezimuel:
Unfortunately, I needed to remove the `docs` folder in `.gitattributes` since it is breaking our internal docs build. I'll investigate more and hopefully revert the change asap.

PR #1390, opened 2024-01-18
Add big `/docs` dir back to export-ignore in gitattributes

bram-pkg:
The directory is quite big, no need to package that along with the SDK when installing using composer


ezimuel:
@bram-pkg unfortunately, I cannot remove this folder in `.gitatttibutes` since the `docs` folder is used by the process that generates the documentation we pages. We are working to change this and in the future this will be possible. Sorry for the inconvenience.

Issue #1398, opened 2024-03-25
[Proposal] Add a mapTo(class) function for map ES|QL response into objects


ezimuel:
This is a proposal for adding a `Response\Elasticsearch::mapTo($class = 'stdClass')` for mapping an ES|QL response into an object of a class `$class`. 

Here an example:
```php
class Employee
{
    public string $name;
    public DateTime $birthdate;
    public float $avg_salary;
    public string $country;
}

$client = ClientBuilder::create()
    ->build();

$result = $client->esql()->query([
    'body' => [
        'query' => 'FROM employees | STATS avg_salary = AVG(salary) by country'
    ]
]);
$employees = $result->mapTo(Employee::class); // array of Employee

printf("Number of employees: %d\n", count($employees));
printf("First employee\n");
printf("Name: %s\n", $employees[0]->name);
printf("Birthdate: %s\n",  $employees[0]->birthday->format('Y-m-d'));
printf("Avg salary: %.2f\n", $employees[0]->avg_salary);
printf("Country: %s\n", $employees[0]->country);

```
If `$class` is not specified the default class is [stdClass](https://www.php.net/manual/en/class.stdclass.php).

ezimuel:
Done in #1399

PR #1399, opened 2024-03-25
Added mapTo feature, see #1398

ezimuel:
As titled, see #1398

### DIFF
--- a/src/Response/Elasticsearch.php
+++ b/src/Response/Elasticsearch.php
@@ -15,17 +15,20 @@ declare(strict_types = 1);
 namespace Elastic\Elasticsearch\Response;
 
 use ArrayAccess;
+use DateTime;
 use Elastic\Elasticsearch\Exception\ArrayAccessException;
 use Elastic\Elasticsearch\Exception\ClientResponseException;
 use Elastic\Elasticsearch\Exception\ServerResponseException;
 use Elastic\Elasticsearch\Traits\MessageResponseTrait;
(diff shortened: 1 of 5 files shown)


src/Response/Elasticsearch.php
```diff
@@ -224,4 +227,63 @@ public function offsetUnset($offset): void
...
+        $iterator = [];
+        $ncol = count($response['columns']);
+        foreach ($response['values'] as $value) {
+            $obj = new $class;
+            for ($i=0; $i < $ncol; $i++) {
+                $field = Utility::formatVariableName($response['columns'][$i]['name']);
+                if ($class !== stdClass::class && !property_exists($obj, $field)) {
+                    continue;
+                }
+                switch($response['columns'][$i]['type']) {
```
pquentin, line 258: Is there a file in Elasticsearch where we can see all the classes? It would be nice to add it in a comment if we have it, in case more types are added in the future.

pquentin (approved):
Thanks! LGTM. Based on my limited knowledge of PHP and without having tested it.

src/Response/Elasticsearch.php
ezimuel: Yes, the list of data types managed by ES|QL is reported here: https://www.elastic.co/guide/en/elasticsearch/reference/current/esql-limitations.html#esql-supported-types

ezimuel:
Thanks @pquentin for the review.

PR #1405, opened 2024-05-27
(Code) docs: Add inline PHPDoc with endpoint var type

gskema:
Added inline PHPDoc with variable type for `$endpoint` so it can be navigated to easily, for example when debugging what request URI is called, etc.


ezimuel (approved):
LGTM. Sorry for the late reply. I missed the notification.

PR #1427, opened 2025-02-20
Move PHPStan settings to it's config

AJenbo:
This makes it so that developers can simply run `vendor/bin/phpstan` and that IDEs can correctly integrate with it for the project.

Also please share you fax number so that I may send you any notices permitted to be given by the CLA.


ezimuel (approved):
LGTM. Thanks!

ezimuel:
Thanks @AJenbo for this PR.

PR #1426, opened 2025-02-20
Fix optional array keys in PHPDocs

AnnaNtagiou:
This fixes #1424 in `Indices` and `ClientEndpointsTrait`

Issue #1424, opened 2025-02-20
Invalid PHPDocs - optional array keys


AnnaNtagiou:
A lot of PHPDocs were invalid as they had comments at the EOL. 

After [PHPStan 2.1.6](https://github.com/phpstan/phpstan/releases/tag/2.1.6), which supports comments at EOL, the PHPDocs are valid but incorrect. There are array keys that are not required but are not defined as optional.

For example, in [ClientEndpointsTrait](https://github.com/elastic/elasticsearch-php/blob/c9a5c85fc886757c6d740e46304665b8c6a988cb/src/Traits/ClientEndpointsTrait.php#L1312-L1323) we have this PHPDoc.

```
         * @param array{
	 *     id: string, // (REQUIRED) Script ID
	 *     context: string, //  Script context
	 *     timeout: time, // Explicit operation timeout
	 *     master_timeout: time, // Specify timeout for connection to master
	 *     pretty: boolean, // Pretty format the returned JSON response. (DEFAULT: false)
	 *     human: boolean, // Return human readable values for statistics. (DEFAULT: true)
	 *     error_trace: boolean, // Include the stack trace of returned errors. (DEFAULT: false)
	 *     source: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
	 *     filter_path: list, // A comma-separated list of filters used to reduce the response.
	 *     body: array, // (REQUIRED) The document
	 * } $params
```

While only the `id` and `body` keys are required, the rest of the keys are not optional.

AJenbo:
To the person looking in to this, optional values should start with a `?`:
```php
       /**
         * @param array{
	 *     id: string, // (REQUIRED) Script ID
	 *     ?context: string, //  Script context
	 *     ?timeout: time, // Explicit operation timeout
	 *     ?master_timeout: time, // Specify timeout for connection to master
	 *     ?pretty: boolean, // Pretty format the returned JSON response. (DEFAULT: false)
	 *     ?human: boolean, // Return human readable values for statistics. (DEFAULT: true)
	 *     ?error_trace: boolean, // Include the stack trace of returned errors. (DEFAULT: false)
	 *     ?source: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
	 *     ?filter_path: list, // A comma-separated list of filters used to reduce the response.
	 *     body: array, // (REQUIRED) The document
	 * } $params
```

[...]

PR #1429, opened 2025-02-22
Update parameter type hints for 'body' in `Indices.php` and `ClientEn…

rajmundtoth0:
This is an addition to  [PR 1426](https://github.com/elastic/elasticsearch-php/pull/1426)

#1424


ezimuel:
@rajmundtoth0 thanks for the PR but unfortunately the `Indices.php` and `ClientEndpointsTrait.php` are generated, you can see the `@generated` tag [here](https://github.com/elastic/elasticsearch-php/blob/main/src/Endpoints/Indices.php#L27) and [here](https://github.com/elastic/elasticsearch-php/blob/main/src/Traits/ClientEndpointsTrait.php#L27). 

We need to apply the changs in the code generator, that is not public. I can apply the changes that you proposed, that is changing the `body: array<mixed>`, right? Thanks again for your contribution and sorry for the late reply.

rajmundtoth0:
@ezimuel yes. No problem, happy to help

Issue #1425, opened 2025-02-20
Invalid type given as required for multiple methods


AJenbo:
https://github.com/elastic/elasticsearch-php/blob/c9a5c85fc886757c6d740e46304665b8c6a988cb/src/Traits/ClientEndpointsTrait.php#L39

`enum` is not a valid PHP type, the correct type for a generic enum is `\UnitEnum`, you probably just want `string`.

`number` is not a valid type, the correct one is `numeric` if you want `int|float|numeric-string`.

`list` is used for comma-separated lists, this should be `string`, but comment might be misleading as it appears it might accept both, in which case `string|list` would be better.

`long` is not a valid type, use `int`

`time` is not a valid type, use `\DateTime`, `int` or `string`

MidnightDesign:
This [has been a issue for almost 2 years](https://github.com/elastic/elasticsearch-php/issues/1346), but nobody is responding.

MidnightDesign:
I think had found it at some point, but I don't know right now.

ezimuel:
@AJenbo, @momala454, @MidnightDesign the code generator is not part of this repository. I'm working to fix the PHPStan issues and I'll let you know asap. Thanks for your patience.

AJenbo:
... right, but what repo is it in?

MidnightDesign:
@ezimuel Thank you for the response. I am aware that it's not here. Where is it?

AJenbo:
For completeness he clarified that the generator is not publicly avalible.

ezimuel:
I added a [CONTRIBUTING](https://github.com/elastic/elasticsearch-php/blob/main/CONTRIBUTING.md) guide where I explained the `@generated` tag. I hope this will help with future contributions. Thanks for your feedback.

ezimuel:
I just sent this PR https://github.com/elastic/elasticsearch-php/pull/1439 to fix the issue

ezimuel:
Just released [v8.17.1](https://github.com/elastic/elasticsearch-php/releases/tag/v8.17.1) with the fix in the PR #1442

Issue #1423, opened 2025-02-20
Invalid phpdoc (list instead of string)


momala454:
In your phpdoc, by example `vendor/elasticsearch/elasticsearch/src/Endpoints/Indices.php`
```php
/**
	 * Updates the index settings.
	 *
	 * @see https://www.elastic.co/guide/en/elasticsearch/reference/master/indices-update-settings.html
	 *
	 * @param array{
	 *     index: list, //  A comma-separated list of index names; use `_all` or empty string to perform the operation on all indices
	 *     master_timeout: time, // Specify timeout for connection to master
	 *     timeout: time, // Explicit operation timeout
	 *     preserve_existing: boolean, // Whether to update existing settings. If set to `true` existing settings on an index remain unchanged, the default is `false`
	 *     reopen: boolean, // Whether to close and reopen the index to apply non-dynamic settings. If set to `true` the indices to which the settings are being applied will be closed temporarily and then reopened in order to apply the changes. The default is `false`
	 *     ignore_unavailable: boolean, // Whether specified concrete indices should be ignored when unavailable (missing or closed)
	 *     allow_no_indices: boolean, // Whether to ignore if a wildcard indices expression resolves into no concrete indices. (This includes `_all` string or when no indices have been specified)
	 *     expand_wildcards: enum, // Whether to expand wildcard expression to concrete indices that are open, closed or both.
	 *     flat_settings: boolean, // Return settings in flat format (default: false)
	 *     pretty: boolean, // Pretty format the returned JSON response. (DEFAULT: false)
	 *     human: boolean, // Return human readable values for statistics. (DEFAULT: true)
	 *     error_trace: boolean, // Include the stack trace of returned errors. (DEFAULT: false)
	 *     source: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
	 *     filter_path: list, // A comma-separated list of filters used to reduce the response.
	 *     body: array, // (REQUIRED) The index settings to be updated
	 * } $params
	 *
	 * @throws NoNodeAvailableException if all the hosts are offline
	 * @throws ClientResponseException if the status code of response is 4xx
	 * @throws ServerResponseException if the status code of response is 5xx
	 *
	 * @return Elasticsearch|Promise
	 */
	public function putSettings(array $params = [])
	{
...
}
```

You use `list` to describe ` comma-separated list of index names`, so a string.
However, a list means an array of numerical indexes starting at 0.

[...]

Issue #1346, opened 2023-08-08
Psalm 5.14.0 and higher triggers issues with defined array-shapes


jaytaph:
Array shape declaration of certain methods are not compatible with psalm 5.14.0+.

```
InvalidArgument - src/Service/Elastic/IndexService.php:114:55 - Argument 1 of Elastic\Elasticsearch\Endpoints\Indices::exists expects array{allow_no_indices: bool, error_trace: bool, expand_wildcards: Elastic\Elasticsearch\Endpoints\enum, filter_path: list<mixed>, flat_settings: bool, human: bool, ignore_unavailable: bool, include_defaults: bool, index: list<mixed>, local: bool, pretty: bool, source: string}, but array{index: string} provided (see https://psalm.dev/004)
        $response = $this->elastic->indices()->exists(['index' => $indexName]);
```

This works correctly in psalm 5.13.1 or lower, but at 5.14.0 there has been some changes that will generate errors when not given all keys found in the definition.

```
	 * @param array{
	 *     index: list, // (REQUIRED) A comma-separated list of index names
	 *     local: boolean, // Return local information, do not retrieve the state from master node (default: false)
	 *     ignore_unavailable: boolean, // Ignore unavailable indexes (default: false)
	 *     allow_no_indices: boolean, // Ignore if a wildcard expression resolves to no concrete indices (default: false)
	 *     expand_wildcards: enum, // Whether wildcard expressions should get expanded to open or closed indices (default: open)
	 *     flat_settings: boolean, // Return settings in flat format (default: false)
	 *     include_defaults: boolean, // Whether to return all default setting for each of the indices.
	 *     pretty: boolean, // Pretty format the returned JSON response. (DEFAULT: false)
	 *     human: boolean, // Return human readable values for statistics. (DEFAULT: true)
	 *     error_trace: boolean, // Include the stack trace of returned errors. (DEFAULT: false)
	 *     source: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
	 *     filter_path: list, // A comma-separated list of filters used to reduce the response.
	 * } $params
```

This will probably function properly by adding a ? after the keyname in the array-shape.

```
	 * @param array{
	 *     index: list, // (REQUIRED) A comma-separated list of index names
	 *     local?: boolean, // Return local information, do not retrieve the state from master node (default: false)
	 *     ignore_unavailable?: boolean, // Ignore unavailable indexes (default: false)
[...]

PR #1439, opened 2025-03-22
Fix for PHPStan issues

ezimuel:
This PR fixes the PHPStan issues reported in #1425, #1424, #1423 and #1346. It also improved the PHPStan rule level to 5.

### DIFF
--- a/src/Endpoints/AsyncSearch.php
+++ b/src/Endpoints/AsyncSearch.php
@@ -35,11 +35,11 @@ class AsyncSearch extends AbstractEndpoint
 	 *
 	 * @param array{
 	 *     id: string, // (REQUIRED) The async search ID
-	 *     pretty: boolean, // Pretty format the returned JSON response. (DEFAULT: false)
-	 *     human: boolean, // Return human readable values for statistics. (DEFAULT: true)
-	 *     error_trace: boolean, // Include the stack trace of returned errors. (DEFAULT: false)
-	 *     source: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
-	 *     filter_path: list, // A comma-separated list of filters used to reduce the response.
+	 *     pretty?: bool, // Pretty format the returned JSON response. (DEFAULT: false)
+	 *     human?: bool, // Return human readable values for statistics. (DEFAULT: true)
+	 *     error_trace?: bool, // Include the stack trace of returned errors. (DEFAULT: false)
+	 *     source?: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
+	 *     filter_path?: string, // A comma-separated list of filters used to reduce the response.
 	 * } $params
 	 *
 	 * @throws MissingParameterException if a required parameter is missing
@@ -49,8 +49,9 @@ class AsyncSearch extends AbstractEndpoint
 	 *
 	 * @return Elasticsearch|Promise
 	 */
-	public function delete(array $params = [])
+	public function delete(?array $params = null)
 	{
+		$params = $params ?? [];
 		$this->checkRequiredParameters(['id'], $params);
 		$url = '/_async_search/' . $this->encode($params['id']);
 		$method = 'DELETE';
@@ -59,7 +60,7 @@ class AsyncSearch extends AbstractEndpoint
 		$headers = [
 			'Accept' => 'application/json',
 		];
-		$request = $this->createRequest($method, $url, $headers, $params['body'] ?? null);
+		$request = $this->createRequest($method, $url, $headers);
 		$request = $this->addOtelAttributes($params, ['id'], $request, 'async_search.delete');
 		return $this->client->sendRequest($request);
 	}
@@ -72,14 +73,14 @@ class AsyncSearch extends AbstractEndpoint
 	 *
 	 * @param array{
 	 *     id: string, // (REQUIRED) The async search ID
-	 *     wait_for_completion_timeout: time, // Specify the time that the request should block waiting for the final response
-	 *     keep_alive: time, // Specify the time interval in which the results (partial or final) for this search will be available
-	 *     typed_keys: boolean, // Specify whether aggregation and suggester names should be prefixed by their respective types in the response
-	 *     pretty: boolean, // Pretty format the returned JSON response. (DEFAULT: false)
-	 *     human: boolean, // Return human readable values for statistics. (DEFAULT: true)
-	 *     error_trace: boolean, // Include the stack trace of returned errors. (DEFAULT: false)
-	 *     source: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
-	 *     filter_path: list, // A comma-separated list of filters used to reduce the response.
+	 *     wait_for_completion_timeout?: int|string, // Specify the time that the request should block waiting for the final response
+	 *     keep_alive?: int|string, // Specify the time interval in which the results (partial or final) for this search will be available
+	 *     typed_keys?: bool, // Specify whether aggregation and suggester names should be prefixed by their respective types in the response
+	 *     pretty?: bool, // Pretty format the returned JSON response. (DEFAULT: false)
+	 *     human?: bool, // Return human readable values for statistics. (DEFAULT: true)
+	 *     error_trace?: bool, // Include the stack trace of returned errors. (DEFAULT: false)
+	 *     source?: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
+	 *     filter_path?: string, // A comma-separated list of filters used to reduce the response.
 	 * } $params
 	 *
 	 * @throws MissingParameterException if a required parameter is missing
@@ -89,8 +90,9 @@ class AsyncSearch extends AbstractEndpoint
 	 *
 	 * @return Elasticsearch|Promise
 	 */
-	public function get(array $params = [])
+	public function get(?array $params = null)
 	{
+		$params = $params ?? [];
 		$this->checkRequiredParameters(['id'], $params);
 		$url = '/_async_search/' . $this->encode($params['id']);
 		$method = 'GET';
@@ -99,7 +101,7 @@ class AsyncSearch extends AbstractEndpoint
 		$headers = [
 			'Accept' => 'application/json',
 		];
-		$request = $this->createRequest($method, $url, $headers, $params['body'] ?? null);
+		$request = $this->createRequest($method, $url, $headers);
 		$request = $this->addOtelAttributes($params, ['id'], $request, 'async_search.get');
 		return $this->client->sendRequest($request);
 	}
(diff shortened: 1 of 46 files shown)


src/Endpoints/AsyncSearch.php
```diff
@@ -35,11 +35,11 @@ class AsyncSearch extends AbstractEndpoint
...
-	 *     pretty: boolean, // Pretty format the returned JSON response. (DEFAULT: false)
-	 *     human: boolean, // Return human readable values for statistics. (DEFAULT: true)
-	 *     error_trace: boolean, // Include the stack trace of returned errors. (DEFAULT: false)
-	 *     source: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
-	 *     filter_path: list, // A comma-separated list of filters used to reduce the response.
+	 *     pretty?: bool, // Pretty format the returned JSON response. (DEFAULT: false)
+	 *     human?: bool, // Return human readable values for statistics. (DEFAULT: true)
+	 *     error_trace?: bool, // Include the stack trace of returned errors. (DEFAULT: false)
+	 *     source?: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
+	 *     filter_path?: string, // A comma-separated list of filters used to reduce the response.
```
AJenbo, line 42: Note the code will actually also accept an array of sting elements: `string|array<string>`

src/Traits/MessageResponseTrait.php
```diff
@@ -29,6 +29,9 @@ public function getProtocolVersion(): string
         return $this->response->getProtocolVersion();
     }
 
+    /**
+     * @return MessageInterface
```
AJenbo, line 33: Nit: This PHPDoc is redundant since it's already expressed by the native return type

src/Traits/ClientEndpointsTrait.php
```diff
@@ -1993,44 +2035,44 @@ public function update(array $params = [])
...
+	 *     expand_wildcards?: string, // Whether to expand wildcard expression to concrete indices that are open, closed or both.
+	 *     lenient?: bool, // Specify whether format-based query failures (such as providing text to a numeric field) should be ignored
+	 *     pipeline?: string, // Ingest pipeline to set on index requests made by this action. (default: none)
+	 *     preference?: string, // Specify the node or shard the operation should be performed on (default: random)
+	 *     q?: string, // Query in the Lucene query string syntax
+	 *     routing?: string, // A comma-separated list of specific routing values
+	 *     scroll?: int|string, // Specify how long a consistent view of the index should be maintained for scrolled search
+	 *     search_type?: string, // Search operation type
+	 *     search_timeout?: int|string, // Explicit timeout for each search request. Defaults to no timeout.
+	 *     max_docs?: float, // Maximum number of documents to process (default: all documents)
```
AJenbo, line 2056: Consider if `int|float` should be the type for `number` as some of these do not appear to make sens as fractions. If it can accept a string representation then you could set it as `numeric` as that is short for `int|float|numeric-string`

src/Traits/ClientEndpointsTrait.php
```diff
@@ -1404,20 +1434,20 @@ public function rankEval(array $params = [])
...
-	 *     body: array, // (REQUIRED) The search definition using the Query DSL and the prototype for the index request.
+	 *     refresh?: bool, // Should the affected indexes be refreshed?
+	 *     timeout?: int|string, // Time each individual bulk request should wait for shards that are unavailable.
+	 *     wait_for_active_shards?: string, // Sets the number of shard copies that must be active before proceeding with the reindex operation. Defaults to 1, meaning the primary shard only. Set to `all` for all shard copies, otherwise set to any non-negative value less than or equal to the total number of copies for the shard (number of replicas + 1)
+	 *     wait_for_completion?: bool, // Should the request should block until the reindex is complete.
+	 *     requests_per_second?: float, // The throttle to set on this request in sub-requests per second. -1 means no throttle.
+	 *     scroll?: int|string, // Control how long to keep the search context alive
+	 *     slices?: float|number|string, // The number of slices this task should be divided into. Defaults to 1, meaning the task isn't sliced into subtasks. Can be set to `auto`.
```
AJenbo, line 1443: This one still has the invalid `number` type as part of it's signature.

It's also valid to have the literal string in the PHPDoc type so the ideal type would be `int|'audo'`

src/Traits/ClientEndpointsTrait.php
```diff
@@ -1311,15 +1339,15 @@ public function ping(array $params = [])
...
-	 *     body: array, // (REQUIRED) The document
+	 *     context?: string, // Script context
+	 *     timeout?: int|string, // Explicit operation timeout
+	 *     master_timeout?: int|string, // Specify timeout for connection to master
+	 *     pretty?: bool, // Pretty format the returned JSON response. (DEFAULT: false)
+	 *     human?: bool, // Return human readable values for statistics. (DEFAULT: true)
+	 *     error_trace?: bool, // Include the stack trace of returned errors. (DEFAULT: false)
+	 *     source?: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
+	 *     filter_path?: string, // A comma-separated list of filters used to reduce the response.
+	 *     body: string|array, // (REQUIRED) The document. If body is a string must be a valid JSON.
```
AJenbo, line 1350: To get closer to passing PHPStan level 6 you might want to hint this as either `string|array<string, mixed>` or `string|array<mixed>` (if lists are allowed)

src/Endpoints/Snapshot.php
```diff
@@ -279,25 +285,25 @@ public function deleteRepository(array $params = [])
...
-	 *     source: string, // The URL-encoded request definition. Useful for libraries that do not accept a request body for non-POST requests.
-	 *     filter_path: list, // A comma-separated list of filters used to reduce the response.
+	 *     snapshot: string, // (REQUIRED) A comma-separated list of snapshot names
+	 *     master_timeout?: int|string, // Explicit operation timeout for connection to master node
+	 *     ignore_unavailable?: bool, // Whether to ignore unavailable snapshots, defaults to false which means a SnapshotMissingException is thrown
+	 *     index_names?: bool, // Whether to include the name of each index in the snapshot. Defaults to true.
+	 *     index_details?: bool, // Whether to include details of each index in the snapshot, if those details are available. Defaults to false.
+	 *     include_repository?: bool, // Whether to include the repository name in the snapshot info. Defaults to true.
+	 *     sort?: string, // Allows setting a sort order for the result. Defaults to start_time
+	 *     size?: integer, // Maximum number of snapshots to return. Defaults to 0 which means return all that match without limit.
```
AJenbo, line 295: Since you changed `boolean` to `bool` you might also want to change `integer` to `int`

src/Endpoints/AsyncSearch.php
ezimuel: @AJenbo these array fields are specificed by the Elasticsearch rest-api-specification, reported [here](https://github.com/elastic/elasticsearch/tree/v8.17.3/rest-api-spec#type-definition). A `list` for this specification is represented as a comma separated list in the querystring, that is equivalent as a PHP string. That's why I used only `string` here.

src/Traits/MessageResponseTrait.php
ezimuel: I agree that this is redundant but if I remove it I get the following error from PHPStan.
```
$ composer run-script phpstan
> phpstan analyse --no-progress --memory-limit 256M
Note: Using configuration file /.../Git/elasticsearch-php/phpstan.neon.
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   Traits/MessageResponseTrait.php (in context of class Elastic\Elasticsearch\Response\Elasticsearch)                                                                                             
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  :34    Method Elastic\Elasticsearch\Response\Elasticsearch::withProtocolVersion() should return static(Elastic\Elasticsearch\Response\Elasticsearch) but returns Psr\Http\Message\ResponseInterface.  
         🪪  return.type                                                                                                                                                                                
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
```
I'm trying to understand where is the issue here, any idea? Thanks.

src/Traits/ClientEndpointsTrait.php
ezimuel: For the Elasticsearch specification reported [here](https://github.com/elastic/elasticsearch/tree/v8.17.3/rest-api-spec#type-definition), a `number` is alias for double (moreover it's deprecated). I agree that makes no sense for `max_docs` that should be always an integer. I will double check with the Elasticsearch team. Thanks!

AJenbo (commented):
Nice work. This change should improve things a lot.

I went over all the changes ad made comments where I think things can still be improved a bit in ways that would make it easier to work with the library and not get false positives.

src/Traits/ClientEndpointsTrait.php
AJenbo: Ok, in that case it sounds like the issue isn't related to this PR but more the upstream documentation 👍

src/Endpoints/AsyncSearch.php
AJenbo: I understand that, but the implementation in this repo will take a list of strings and turn it in to a comma separated string:
https://github.com/elastic/elasticsearch-php/blob/603d87ed95fe47eee41bfaa32f855d38e8a89b80/src/Traits/EndpointTrait.php#L71

Not documenting this effectively turns this functionality in to dead code and prevents us from taking advantage of the convenience.

src/Traits/MessageResponseTrait.php
AJenbo: This looks to be because Elasticsearch is actually violating the ResponseInterface interface by returning an instance of the underlying ResponseInterface rather then the object it's being called on.

https://github.com/php-fig/http-message/blob/402d35bcb92c70c026d1a6a9883f06b2ead23d71/src/MessageInterface.php#L39

By adding the PHPDoc to the trait it's overwriting the one from the ResponseInterface which is why it silences the error.

I have opened a PR to fix this properly: https://github.com/elastic/elasticsearch-php/pull/1440

src/Endpoints/AsyncSearch.php
ezimuel: @AJenbo unfortunately the `convertValue()` function that is able to translate an array<string> into a comma separated value is used only for the query string parameter (e.g. `filter_path`). I need to adjust the usage of `convertValue` also for other parameters in the URL, e.g. [here](https://github.com/elastic/elasticsearch-php/blob/main/src/Endpoints/AsyncSearch.php#L152). Anyway, you right we should use this feature with `string|array<string>`. Thanks for you help!

src/Traits/MessageResponseTrait.php
ezimuel: Thanks @AJenbo nice catch! I'll ask to wait the #1440 until we merge this PR: Moreover, can you rebase the PR to 8.17 branch? Thanks.

ezimuel:
Thanks @AJenbo for all your feedback and contribution. I'm finalizing this PR with all your feedback.

src/Traits/MessageResponseTrait.php
AJenbo: Done, I also set it to target the 8.17 branch so you can make it part of this PR.

ezimuel:
I'm closing this PR since it has been sent to main by mistake. I need to revert some changes and send it to 8.17.

PR #1440, opened 2025-03-22
Fix instance returned by MessageResponseTrait

AJenbo:
The PSR-7 ResponseInterface states that the method here should return an instance of the same class as the method is being called on but the trait instead returned an instance of the underlying implementation. Simply returning `$this` resolves the issue and likely is also what is intended by the ResponseInterface having a return type of `static`.

*Note that the failing tests here are the same as on the main branch and not related to the changes*

### DIFF
--- a/src/Traits/MessageResponseTrait.php
+++ b/src/Traits/MessageResponseTrait.php
@@ -29,12 +29,11 @@ trait MessageResponseTrait
         return $this->response->getProtocolVersion();
     }
 
-    /**
-     * @return MessageInterface
-     */
     public function withProtocolVersion($version): MessageInterface
     {
-        return $this->response->withProtocolVersion($version);
+        $this->response->withProtocolVersion($version);
+
+        return $this;
     }
 
     public function getHeaders(): array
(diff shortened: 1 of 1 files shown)


src/Traits/MessageResponseTrait.php
```diff
@@ -29,12 +29,11 @@ public function getProtocolVersion(): string
         return $this->response->getProtocolVersion();
     }
 
-    /**
-     * @return MessageInterface
-     */
     public function withProtocolVersion($version): MessageInterface
     {
-        return $this->response->withProtocolVersion($version);
+        $this->response->withProtocolVersion($version);
```
ezimuel, line 34: We need to return the `withProtocolVersion($version)` otherwise we'll lost the response. We cannot return $this, the [PSR-7](https://www.php-fig.org/psr/psr-7/) MessageInterface is immutable.

ezimuel (commented):
This PR from a PHPStan point of view is right but it doesn't work properly since the various `withProtocolVersion()`, `withHeader()`, etc do not return values. Thanks anyway for the contribution, you helped me to solve most of the PHPStan issues in #1442

src/Traits/MessageResponseTrait.php
AJenbo: In that case you need to do a bigger refactoring in order to resolve this properly. withProtocolVersion() should not return a different implementation of MessageInterface

AJenbo: maybe a good solution would be to clone the object and swap the response object with the new one and then return that.

ezimuel: I think this is more a PHPStan issue rather that a real PHP one. In the meantime we'll leave the extra comment to prevent the error message from PHPStan.

PR #1442, opened 2025-03-25
Fix and improvements for PHPStan (rule level 5)

ezimuel:
This PR replaces #1439. It includes all the feedback from @Ajenbo given in #1439. Thanks @AJenbo!

It fixes the PHPStan issues reported in https://github.com/elastic/elasticsearch-php/issues/1425, https://github.com/elastic/elasticsearch-php/issues/1424, https://github.com/elastic/elasticsearch-php/issues/1423 and https://github.com/elastic/elasticsearch-php/issues/1346. It also improved the PHPStan rule level to 5 :tada:

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,6 +1,4 @@
 parameters:
-    level: 2
+    level: 5
     paths:
         - src
-    ignoreErrors:
-        - '#PHPDoc tag @param has invalid value#'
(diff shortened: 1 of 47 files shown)


AJenbo:
Looks good, very excited for this to be released so we can start using it :)

Looks like after this as been merged getting to PHPStan lvl 6 will also be doable which would mean that all interfaces are documented.

miguelgrinberg (approved):

AJenbo:
What's the process for this ending up in a release? I see it's merged into a version branch and not the latest?

PR #1462, opened 2025-09-10
ESQL query builder prototype

miguelgrinberg:
This PR contains a prototype of an ES|QL query builder feature for the PHP client.

Example usage:

```php
<?php

require(__DIR__ . "/vendor/autoload.php");

use Elastic\Elasticsearch\ClientBuilder;
use Elastic\Elasticsearch\Helper\ESQL;

    ->setHosts([getenv("ELASTICSEARCH_URL")])
    ->setApiKey(getenv("ELASTIC_API_KEY"))
    ->build();

$query = ESQL::from("books", "books*")
    ->where('author == "King"', 'year == 1982');
echo $query;

$result = $client->esql()->query([
    'body' => ['query' => $query->__toString()]
]);
```

The output of `echo $query` for the code above is:

```
FROM books, books*
| WHERE author == "King" AND year == 1982
```

### DIFF
--- a/src/Helper/Esql.php
+++ b/src/Helper/Esql.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * Elasticsearch PHP Client
+ *
+ * @link      https://github.com/elastic/elasticsearch-php
+ * @copyright Copyright (c) Elasticsearch B.V (https://www.elastic.co)
+ * @license   https://opensource.org/licenses/MIT MIT License
+ *
+ * Licensed to Elasticsearch B.V under one or more agreements.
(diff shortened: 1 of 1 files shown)


miguelgrinberg:
@ezimuel This is ready to review!

ezimuel (approved):
Thanks @miguelgrinberg for this awesome PR! I just left a couple of comments, the most relevant that I would suggest to investigate is the possibility to use Traits instead of passing the parent in EsqlBase.
Moreover, we should add also some documentation about this helper with an example on how to call it using the Esql endpoint.

PR #1467, opened 2025-09-15
Fix invalid ip address for 9.0 branch

ketkee-aryamane:
Linked to https://github.com/elastic/docs-content-internal/issues/153
Fixed invalid IP Addresses per RFC: https://datatracker.ietf.org/doc/html/rfc5737


ezimuel (approved):
LGTM

PR #1473, opened 2025-10-27
Clear cached response data when resetting Response object

zigzagdev:
### Description

This pull request introduces a small but important fix to ensure response consistency across multiple calls to setResponse() within the Elastic\Elasticsearch\Response\Elasticsearch class.

Previously, when a new PSR-7 ResponseInterface was passed to setResponse(), any previously cached data in $asArray, $asObject, or $asString persisted. This could lead to incorrect data being returned if the same Elasticsearch instance was reused.

To resolve this, the following change has been applied:

```php

unset($this->asArray, $this->asObject);
$this->asString = '';

```
These lines ensure that any previously serialised body representations are cleared whenever a new response is set.

### Why

Without clearing the cached serialised data, the same `Elasticsearch` instance could return
stale results from a previous response. This becomes problematic when multiple responses
are handled in sequence — for example, in integration tests or long-running processes that
reuse the same client instance.

Resetting these cached properties ensures that the object always reflects the latest
response body, maintaining consistency and correctness without affecting public APIs.
This change is small, isolated, and backward compatible, yet it prevents subtle and
potentially confusing data inconsistencies.

### Tests Results
スクリーンショット 2025-10-27 14 18 06

I hope this fix will help for !!! 🙌


zigzagdev:
@miguelgrinberg 

> This PR includes several formatting changes that are unrelated to the fix in question. I would prefer if those were removed and if necessary, submitted separately as another improvement.

Thanks for pointing that out !
You're right. And those formatting changes are unrelated to the fix.
I'll revert them from this PR and submit them separately if needed.

PR #1560, opened 2026-03-20
Bulk helper

### LINKED ISSUE #1537: Bulk helper
Other language clients offer a bulk helper that makes it easier and more idiomatic to ingest using the bulk endpoint. This task is about adding a similar feature to the PHP client.

miguelgrinberg:
Closes #1537 

A bulk helper class that creates and submits chunk requests according to size or operation count limits. The application just provides an iterator or generator that feeds items one by one. Example:

```php
use Elastic\Elasticsearch\Helper\Bulk;

function get_next_document() {
    foreach($dataset as $document) {
        yield Bulk::index_action($document)
    }
}

Bulk::bulk($client, $index, get_next_document());
```


.github/workflows/yaml_test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         os: [ubuntu-latest]
         es-version: [9.5.0-SNAPSHOT]
         test-suite: [stack]
-        branch-client-tests: ["main"]
+        branch-client-tests: ["9.4"]
```
picandocodigo: This needs to remain being `main` for the main branch, the script will bump it when we create the release branch, but [only if it's not main](https://github.com/elastic/elasticsearch-php/blob/main/.github/make.sh#L177).

miguelgrinberg: I changed it because the tests fail with main. There are some endpoints that were feature flag in 9.4 and are not anymore in main.

miguelgrinberg: I reverted this now, this isn't needed anymore. Thanks!

